PR #10, opened 2020-02-22
Remove CuPy v7.2.0 build 0

leofang:
We just merged conda-forge/cupy-feedstock#39 and the updated packages are up online, so the buggy ones should be marked as broken.

cc: @jakirkham


jakirkham:
Thanks Leo! 😄

jakirkham:
Looks like they are marked now 😄

Screen Shot 2020-02-21 at 8 47 59 PM

PR #12, opened 2020-02-27
Mark abseil-cpp-20200225 as broken

xhochy:
See https://github.com/conda-forge/abseil-cpp-feedstock/issues/8

cc @pitrou


pitrou:
I don't know how this works so I'm enable to give an opinion on the PR.

xhochy:
@conda-forge/core Please give this a review/merge.

bluescarni:
From the error logs posted elsewhere, it looks like the abseil CMake config file machinery is hard-coding paths at build time. There was a similar problem with frameworks under OSX which I hacked around with a patch in the recipe. Perhaps we should start raising these issue upstream.

PR #15, opened 2020-03-05
Pull Python 3.6 packages with ABI tags

jakirkham:
It appears the PyPy packages are being pulled in when CPython is expected. This appears to be causing builds to break. For now, let's pull these packages until someone is able to debug this further.

xref: https://github.com/conda-forge/conda-forge.github.io/issues/867#issuecomment-594985046

cc @conda-forge/core


jakirkham:
Thanks @isuruf! 😀

PR #17, opened 2020-03-11
Remove the openmpi package with CUDA-awareness enabled

leofang:
In conda-forge/openmpi-feedstock#56 we temporarily disabled (but still built) the support for CUDA awareness. This is to remove the package that enabled the support by default.

cc: @jakirkham @dalcinl


jakirkham:
Thanks Leo! 😄

leofang:
Thanks!

PR #21, opened 2020-03-18
Mark broken PyDM 1.9.0 build 0.

hhslepicka:


ocefpaf:
@hhslepicka just for future reference, what is the reason for moving this to broken?

hhslepicka:
I missed the entry_points section at the yaml file and the package generated a wrong script

PR #23, opened 2020-03-19
Mark Spyder 4.1.1 build 0 as broken

ccordoba12:
This package has the wrong Jedi dependency. See https://github.com/conda-forge/spyder-feedstock/pull/79 for the fix.

Build 1 is already up with the fix.


ccordoba12:
@conda-forge/core, could you merge this one? Thanks!

PR #28, opened 2020-03-24
Remove some CuPy v7.2.0 packages

leofang:
Not sure if we can leave comments in the txt file, so let me explain here:

- The first two batches are PyPy builds (build numbers 2 & 3 x CUDA versions 9.2, 10.0, 10.1, 10.2). They should be marked broken because of the observation https://github.com/conda-forge/cupy-feedstock/issues/43#issuecomment-603032749.
- The rest two batches are CUDA 10.2 builds (build number 1 & 2 x CPython 3.6, 3.7, 3.8). They are broken because the fp16 headers are missed (conda-forge/cupy-feedstock#41).

@jakirkham, could you confirm the list is correct? Thanks!

### DIFF
--- a/pkgs/cupy.txt
+++ b/pkgs/cupy.txt
@@ -1,18 +1,21 @@
+# PyPy builds (build no.2)
 linux-64/cupy-7.2.0-py36hf411039_2.tar.bz2
 linux-64/cupy-7.2.0-py36he7af1ec_2.tar.bz2
 linux-64/cupy-7.2.0-py36hbfcebc9_2.tar.bz2
 linux-64/cupy-7.2.0-py36h6e251b3_2.tar.bz2
 
+# PyPy builds (build no.3)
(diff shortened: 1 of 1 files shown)


leofang:
Looks like we can't even have blank lines in the list...Let me modify the file and the PR description.

isuruf:
Or fix the code to ignore blank lines and lines starting with `#`. :smile:

isuruf:
nvm, we are deleting that file, so not worth the effort.

leofang:
Oh, you meant `mark_broken.py`? I was just about to write a PR 🙂

leofang:
https://github.com/conda-forge/cf-mark-broken/compare/master...leofang:ignore_lines

isuruf:
Looks good. You also need to fix Line 24

leofang:
#29 works.

jakirkham:
Thanks Leo! Also thanks Isuru for the review! 😄

PR #36, opened 2020-04-08
Remove Open MPI v4.0.3 for OS X

leofang:
On OS X we learned that there's a cleanup failure in the fresh v4.0.3, see https://github.com/conda-forge/openmpi-feedstock/pull/57#issuecomment-610678486. We are still discussing what to do for OS X, and in the meanwhile it's best to remove the OS X build.

cc: @dalcinl


leofang:
ping @conda-forge/core

leofang:
Thanks, Matt!

PR #38, opened 2020-04-12
Fix CODEOWNERS

xhochy:
@conda-forge/core


xhochy:
Codeowners should be case insensitive. See also https://github.com/conda-forge/r-arrow-feedstock/blob/master/.github/CODEOWNERS for a working group example

PR #41, opened 2020-04-23
ENH mark osx build of python broken due to extra files

beckermr:
cc @conda-forge/python 

xref: https://github.com/conda-forge/python-feedstock/issues/339


pkgs/python_osx.txt
```diff
@@ -1,2 +1,2 @@
 osx-64/python-3.6.10-h4334963_1010_cpython.tar.bz2
-osx-64/ython-3.8.2-hd5f0129_6_cpython.tar.bz2
```
beckermr: 🤦

jjhelmus:
I approve of this.  Getting to the bottom of the problem with these builds will likely take some time and is not something I can look into at this time.

PR #43, opened 2020-04-27
Mark python-language-server versions less than 0.31.10 as broken

ccordoba12:
- They don't have the right upper dependency on Pyflakes, which created a conflict with flake8.
- This was fixed in PR conda-forge/python-language-server-feedstock#50. Packages with the fix are in anaconda.org already.


ccordoba12:
Pinging @jjhelmus about this one so they can apply a similar fix on the Anaconda side.

ccordoba12:
@conda-forge/core, could you take a look at this one? Thanks!

isuruf:
As this is removing full versions, they will not be installable. Should this be a repodata hotfix instead?

jjhelmus:
A repodata hotfix to adjust the existing packages would be preferred.  I will need to create one for `defaults` as `python-language-server` is used in the `anaconda` metapackage.  Removing the packages from `defaults` would leave the metapackage un-installable.

ccordoba12:
> As this is removing full versions, they will not be installable. Should this be a repodata hotfix instead?

I thought this was the accepted procedure when maintainers detect previous versions with incorrect dependencies (at least that's what I've done in previous PRs here).

So what do I have to do to make a repodata hotfix?

ccordoba12:
Pinging @conda-forge/core again: what do I need to do to fix this problem?

jjhelmus:
@ccordoba12 A PR to https://github.com/conda-forge/conda-forge-repodata-patches-feedstock that adjusts the `pyflakes` requirement for the effected versions of `python-language-server` would be the best option.

jjhelmus:
@ccordoba12 conda-forge/conda-forge-repodata-patches-feedstock#45 patches the repodata for the existing packages to add the correct pylint upper bound to `python-language-server`.  Could you check that the limits in that PR are correct?

ccordoba12:
Closing this one then. Thanks for your help @jjhelmus!

PR #45, opened 2020-05-03
mark  osx-64/freeimage-3.18.0-h4a8c4bd_0 as broken

looooo:
we encountered some linker issues with occt. Somehow the older build is used at runtime while the newer package is used during the build.
https://github.com/conda-forge/occt-feedstock/issues/42


isuruf:
cc @conda-forge/freeimage

wolfv:
:+1:

PR #54, opened 2020-05-16
Mark recent OpenJDK builds as broken

xhochy:
These break `rjava` by putting `libjvm` into a non-standard location: https://github.com/conda-forge/openjdk-feedstock/issues/66

Note that for openjdk11, build 1019 did not run on Azure, so there are no builds for linux-/win-/osx-64 to be marked as broken.

Checklist:

      ping @conda-forge/openjdk @CJ-Wright 

PRs to fix the build:

- https://github.com/conda-forge/openjdk-feedstock/pull/70
- https://github.com/conda-forge/openjdk-feedstock/pull/69

### DIFF
--- a/pkgs/openjdk.txt
+++ b/pkgs/openjdk.txt
@@ -1,0 +1,5 @@
+osx-64/openjdk-8.0.192-h0b31af3_1004.tar.bz2
+win-64/openjdk-8.0.192-1004.tar.bz2
+linux-64/openjdk-8.0.192-h516909a_1004.tar.bz2
+linux-ppc64le/openjdk-11.0.1-h08ffdf3_1019.tar.bz2
+linux-aarch64/openjdk-11.0.1-h600c080_1019.tar.bz2


CJ-Wright:
@isuruf can you expand on why?

isuruf:
That was added to fix another package and it's generally good to have all the libraries under `$PREFIX/lib`. `r-rjava` should be fixed to find `libjvm.so` at the correct place

xhochy:
> That was added to fix another package and it's generally good to have all the libraries under `$PREFIX/lib`. `r-rjava` should be fixed to find `libjvm.so` at the correct place

The correct place is where has been located all the time. It is not expected to be placed in directly lib/. Otherwise we will have to patch even more packages that depend on Java in the future.

xhochy:
Further notes:
- The source of the symlink was a false positive of a linter, not an actual issue wit the package.
- The package that would need otherwise be fixed is r-base, not r-rjava.

isuruf:
I looked in a few other distributions and `libjvm.so` is not in `$PREFIX/lib` in any of them.

marcelm:
Hi, we get CI failures now because we were pinning `openjdk==8.0.192=h516909a_1004`. We don’t have R in our environment, so we weren’t affected by any problems in relation to that.

For us, the "damage" done by pulling the package off the channel was the actual problem. Perhaps marking something as broken should be reserved for more serious cases?

beckermr:
What are you pinning to the build? That is generally a mistake.

marcelm:
I’m using the output of `conda env export`. I’m aware there’s `--no-builds`, but the [docs for sharing environments](https://docs.conda.io/projects/conda/en/latest/user-guide/tasks/manage-environments.html#sharing-an-environment) don’t include this as part of the recommended command line.

Perhaps you can convince me otherwise, but I do think that pinning down to the revision is the right thing to do for reproducible tests.

I have a second, much shorter, hand-written `environment.yml` file that only lists direct dependencies with mostly unpinned requirements. The `conda env export` environment is generated from that.

xhochy:
Note that we are very regulary marking builds as broken, this repository is only a subset of the packages that were marked as broken. If you want to depend on builds, either include the broken channel in your install (but not your requirements pinning) or better use `conda-pack` to persist the built environment.

The breakage here has popped up in R but could also affect other JVM-related packages that use `libjvm.so` for detecting the location of the JDK.

beckermr:
As @xhochy said, depending on precise build numbers when you don't need to is going to lead to breakages. 

I would **not** treat that part of the documentation as a recommended approach but more as a general how-to. It is much better, in general, to use semantic pinnings unless it is absolutely necessary to pin otherwise

marcelm:
Thanks, that gives me a couple of options! Conda-lock is new to me, it looks interesting – although it does say this:
> Use cases: [...] Archiving an environment in a functioning state. Note that a more sustainable way to do this is to specify your environment as a environment.yml, and recreate the environment when needed.

I’ll also reconsider using `--no-builds`.

Perhaps the lesson to learn from this is that the story around persisting environments isn’t as clear as it should be.

marcelm:
@beckermr It’s the question of using locked vs abstract dependencies, and both are valid approaches that have different use cases. The semantic/abstract pins are good for development, but the concrete/locked ones are good on CI, where I don’t want breakage just because a dependency was updated (I want to find those breakages of course, but I want to choose when). There’s a discusson of "lock files for Conda" at https://github.com/conda/conda/issues/4205 , and my approach kind of follows that idea.

beckermr:
> Perhaps the lesson to learn from this is that the story around persisting environments isn’t as clear as it should be.

Yes. How one persists an env is a function very much a function of your intent for the end-user. There is no one-size-fits-all approach.

isuruf:
After https://github.com/conda-forge/openjdk-feedstock/pull/69 is merged, we can also undo this. It's not a huge problem to have this lying around.

PR #55, opened 2020-05-18
ENH move labels and trigger build

beckermr:
cc @isuruf @brianv0

### DIFF
--- a/mark_broken.py
+++ b/mark_broken.py
@@ -2,6 +2,8 @@ import sys
 from glob import glob
 import subprocess
 import os
+import tempfile
+
 
 def split_pkg(pkg):
     if not pkg.endswith(".tar.bz2"):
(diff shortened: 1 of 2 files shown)


mark_broken.py
```diff
@@ -65,6 +77,34 @@ def mark_broken():
...
+            shell=True
+        )
+
+        subprocess.check_call(
+            "git commit --allow-empty -am 'resync repo data for broken packages'",
+            cwd=os.path.join(tmpdir, "conda-forge-repodata-patches-feedstock"),
+            shell=True
+        )
+
+        subprocess.check_call(
```
isuruf, line 102: If check call fails, the token will be printed to stdout right?

beckermr: On many ci services the tokens get masked by the ci service itself. GitHub redacts secrets in logs.

beckermr: Quote from the docs 

> To ensure that GitHub redacts your secret in logs, avoid using structured data as the values of secrets. For example, avoid creating secrets that contain JSON or encoded Git blobs.

beckermr:
This will break if we need a rerender FWIW. Not a show stopper, but something to be aware of.

isuruf:
@beckermr, can you describe the new way for @conda-forge/core? We can feedback then.

beckermr:
The new way for what?

isuruf:
I mean `main` -> `main, broken` instead of `main`->`broken`, motivation for this change and caveats.

beckermr:
I guess you mean this from gitter:

> Here's an idea. Instead of removing the label main in anaconda.org and adding broken in anaconda.org.
We could do the following,
1. Add both main and broken in anaconda.org
2. Mark the broken packages as a removal in conda-forge-repodata-patches
This way, the full URL still works, but the package is not in repodata.json so conda solver will not see it

beckermr:
Right, the caveats are that this leaves inconsistent and sometimes weird repodata lying around. Particularly, the repodata on conda-web will now be **wrong** since it doesn't have the patches needed to remove the broken stuff.

beckermr:
Another caveat is from @jjhelmus 

> One caveat is that I think hotfixing only applied to the main label on Anaconda.org. So the package requirements would appear different between the main and broken labels.

beckermr:
The end result of all of this is that envs with packages that get marked broken later can be recreated since the actual packages will still get pushed to the CDN. New envs won't use them since they are removed from the repodata.

beckermr:
We can merge this one in about an hour @isuruf

beckermr:
Ready to merge @isuruf @conda-forge/core ?

PR #78, opened 2020-06-10
Mark message-ix arch-specific files as broken

khaeru:
Users of [message_ix](https://github.com/iiasa/message_ix) report, and we have confirmed, that the presence of these arch-specific files makes `conda` want to install them, rather than the latest version, which is noarch. 

After wrestling with `conda install`, I have been unable to devise a combination of options/configuration/channel priority that makes the solver follow user intent. Since these versions are old and no longer supported, I am hoping that removing them from the repository will prevent the solver from being confused.

Checklist:
* [x] Added a link to the relevant issue in the PR description. (Above.)

### DIFF
--- a/pkgs/message-ix.txt
+++ b/pkgs/message-ix.txt
@@ -1,0 +1,6 @@
+linux-64/message-ix-1.0.0-2.tar.bz2
+linux-64/message-ix-1.0.0-3.tar.bz2
+osx-64/message-ix-1.0.0-2.tar.bz2
+osx-64/message-ix-1.0.0-3.tar.bz2
+win-64/message-ix-1.0.0-2.tar.bz2
+win-64/message-ix-1.0.0-3.tar.bz2


xhochy:
This is probably https://github.com/conda/conda/issues/9905 and thus should be fixed in `conda` itself instead of us starting to mark old but working packages as broken.

xhochy:
Workarounds that are known:

- Use `CONDA_REPODATA_FNS=repodata.json conda …`
- Use `mamba` instead of `conda`

The only viable solution is to fix this in conda itself as we see many other users run into this.

khaeru:
Thanks for the extra information.

Most of our users are entirely new to Anaconda; we suggest it as an install route to avoid pitfalls installing some heavier-weight dependencies like OpenJDK. When conda works, it reduces opportunities for user error, however, even then these beginner users need a significant amount of tech support.

The workarounds you suggest may work, but:
- they introduce additional complexity that we then need to communicate to our users, thus negating any benefit of using conda in the first place.
- the linked issue has been open for more than a month, and we need to resolve this issue somewhat urgently. There doesn't appear to be a clear statement of when it will be resolved, or how.
- mamba labels itself as experimental.

No user is working with the 2½-year-old versions listed in this PR. We no longer support them. Nothing else depends on them. If anyone should need to use them, it will be an expert user doing some kind of archival work, so they will be able to access old code via PyPI and GitHub. So the impact of this removal is nil.

I agree conda *should* fix the upstream issue, and it *should* not be necessary to remove these files. But, for the above reasons, this is our preferred solution.

Please merge.

isuruf:
I tried creating a new environment and I get the latest version.

```
conda create -n testenv message-ix
```

The issue you linked above uses `conda install -c conda-forge message-ix` and we don't support adding conda-forge just for an install line. You should communicate to your users that the correct installation instructions are at https://github.com/conda-forge/message-ix-feedstock#installing-message-ix

khaeru:
@isuruf, thanks for weighing in.

As I said above, we have spent quite some time investigating this, in search of fixes.

iiasa/message_ix#358 shows [two GitHub Actions workflows](https://github.com/iiasa/message_ix/pull/358/commits/89660ac4456cc4dc28e237da43030e6d6b588b40); [here](https://github.com/iiasa/message_ix/pull/358/checks?check_run_id=764568915) is one that succeeds with Miniconda, and [here](https://github.com/iiasa/message_ix/pull/358/checks?check_run_id=764568889) is a failing one with Anaconda. Notice in the latter that the commands are exactly as you say, and that 1.0.0-3 is installed.

So, perhaps you're using Miniconda, or your system is in some other state; I don't know. But it is repeatable on a completely clean install of Anaconda. A fix will be repeatable if this check passes.

Some options I see here:
1. conda-forge can correct the auto-generated install instructions in the feedstock README, which do not work; e.g. by specifying conditions under which they actually will work.
2. If you have alternate instructions other than the non-working ones, I can try plugging them into the GitHub Actions workflow. Again, we have already tried about a dozen different combinations, and the only observed outcomes are either (a) installing the outdated version or (b) the solver fails after a long time. But perhaps conda's documentation is just not clearly pointing to the solution here.
3. We can merge this PR.

isuruf:
@khaeru, we don't support installing in an anaconda environment. Anaconda is a distribution with strict package pins and I don't think that this would help. If you are using anaconda, you have to create a new environment.

isuruf:
As https://github.com/iiasa/message_ix/pull/358#issuecomment-642576993 mentions even with explicitly installing 3.0.1, you still get an unsatisfiable error. This is because of the `anaconda` meta package in the anaconda distribution which has strict pins.

ocefpaf:
I guess we can close this one. Please re-open if that is not the case.

khaeru:
> we don't support installing in an anaconda environment. Anaconda is a distribution with strict package pins and I don't think that this would help. If you are using anaconda, you have to create a new environment.

> This is because of the `anaconda` meta package in the anaconda distribution which has strict pins.

@isuruf, thanks for clarifying those points.

As I mentioned above, this is not made adequately clear by the documentation, in at least two places (below). Can you please let me know where to open (an) issue(s) requesting that the documentation be fixed?

1. The main conda-forge docs at https://conda-forge.org/docs/user/introduction.html#how-can-i-install-packages-from-conda-forge contain this note:

   > In addition to the channel priority, we recommend to always install your packages inside a new environment instead of the base (formerly known as root) environment, and we also recommend the use miniconda instead of the Anaconda Distribution. Using environments make it easier to debug problems with packages and ensure the stability of your base environment. Avoiding the Anaconda Distribution reduces chances of unsolvable/conflicts installations, it is also a smaller download.

   This should be more explicit, as you have been in your comments here: about the strict package pins, and to say "we don't support"/"this will not work" instead of "we recommend", which is not the same.

2. The template README in the feedstock repo does not contain any mention of this limitation or a link to the documentation in (1) above: https://github.com/conda-forge/ixmp-feedstock#installing-ixmp.

PR #80, opened 2020-06-12
Remove `pyarrow` packages restricted to CUDA 9.2

jakirkham:
These packages were missing `ignore_run_exports`, which they should have had as they do not need a specific CUDA version (unlike what is typically the case). Instead they build against the oldest supported CUDA version and make that the lower bound. However as `ignore_run_exports` was not applied to `cudatoolkit`, the result was they were pinned to CUDA 9.2, which made it impossible to install them with newer CUDA versions (as was intended). This has since been fixed. However we should mark these packages with the incorrect constraint as `broken` to avoid them getting pulled in.

cc @conda-forge/pyarrow @conda-forge/arrow-cpp

Checklist:

xref: https://github.com/conda-forge/arrow-cpp-feedstock/issues/149
xref: https://github.com/conda-forge/arrow-cpp-feedstock/pull/150

### DIFF
--- a/pkgs/pyarrow_cuda_9.2.txt
+++ b/pkgs/pyarrow_cuda_9.2.txt
@@ -1,0 +1,3 @@
+linux-64/pyarrow-0.17.1-py36h1234567_3_cuda.tar.bz2
+linux-64/pyarrow-0.17.1-py37h1234567_3_cuda.tar.bz2
+linux-64/pyarrow-0.17.1-py38h1234567_3_cuda.tar.bz2


isuruf:
This should not have been merged. These packages are not broken. They have a stricter pin and stricter pins are fine to leave in the metadata

jakirkham:
These were causing problems for us and we’ve agreed that they needed to be fixed (and since have). I’m not sure I understand the issue with removing them.

isuruf:
What problems? Who agreed?

jakirkham:
The issue and PR are linked above

isuruf:
We don't remove packages that are not broken, just because some person find it convenient to mark them broken. We mark them broken, if they are actually broken.

isuruf:
The issue and the PR link above shows that the new package has the "correct" dependencies. It doesn't mean that the older builds have incorrect dependencies.

For eg: when we make a package noarch, do we mark the non-arch packages as broken? No, because they are okay to have.

isuruf:
@scopatz, can you please revert this until there's consensus?

pearu:
> However we should mark these packages with the incorrect constraint as `broken` to avoid them getting pulled in.

In what situations these packages (build=3) would be pulled in considering that a new build (=4) is available? If none, there is no issue to fix here.

A general consideration: new builds of packages often fix packaging issues of various kinds. What would be the criteria to mark the older builds as broken? My understanding is that marking older builds broken is not practiced in general unless these still cause issues when newer builds are available.

isuruf:
> If none, there is no issue to fix here.

I think there's none. @jakirkham can you respond?

> My understanding is that marking older builds broken is not practiced in general unless these still cause issues when newer builds are available.

Agreed.

@scopatz, can you please revert this?

beckermr:
> @scopatz, can you please revert this?

Reverting won't work on its own FYI. Someone will have to go and remove labels by hand.

isuruf:
Yes, that's what I'm asking to do.

jakirkham:
I'm not understanding why we want to revert this. Do you need these packages for some reason @isuruf?

The packages were overly constrained. I put the PR up as people were still reporting issues to me. So I wanted to make sure these weren't getting pulled in. Unless there is a compelling need for the packages I would recommend leaving them as `broken`.

isuruf:
Can you please respond to @pearu's comment?

> I put the PR up as people were still reporting issues to me. 

Please stop spreading FUD. If you have a specific issue, please report it. I highly doubt that after `build=4` is installed that `build=3` causes any problems.

My main concern is allowing PRs as such as this marking packages that are not broken to be marked broken.

isuruf:
> Do you need these packages for some reason @isuruf?

I haven't used `arrow` or know what it is. I just care about the conda-forge eco-system.

xhochy:
(With my conda-forge hat on, especially not as a `pyarrow` maintainer)

We seem to have a tendendency to mark things quickly as broken when things already can be fixed through a `conda update` (which is necessary either way). This hinders people to reproduce installs quite a lot and also makes the work of the solver unnecessarily complex.

beckermr:
I agree with @xhochy. We have solved some of the reproducibility issues but too many remain.

jakirkham:
I'm sorry, but I can't give more details as the problems were being reported to me by colleagues on Slack last night. So I lack the details to reproduce myself and lack a link that can be easily shared.

isuruf:
Ping on this @scopatz. It seems two pyarrow maintainers are in favour of reverting this.

isuruf:
@scopatz, you merged this PR in less than 15 minutes, but has been silent for 4 days. What's going on?

isuruf:
@conda-forge/core, ping on this.

jakirkham:
@isuruf, can I ask that you please add this to the meeting for next week?

isuruf:
We already discussed this in a meeting on April 29. @jjhelmus proposed that we don't mark any packages that have metadata wrong as broken and instead hotfix the metadata and we all agreed. You say that the metadata is wrong and then marked it broken. (I even question the assumption that the metadata is wrong, because overly restrictive metadata is not wrong if there's a new build)

isuruf:
Ping on this again, @conda-forge/core

isuruf:
@scopatz gave the green light to revert this. I'll revert this tomorrow.

PR #85, opened 2020-06-12
mark grpc packages as broken

s22chan:
they do not have the correct grpc version pinning dependency

See https://github.com/conda-forge/google-api-core-feedstock/issues/44

Checklist:

<!--

      ping @conda-forge/foo
--!>
ping @conda-forge/google-api-core-grpc

### DIFF
--- a/pkgs/google-api-core-grpc.txt
+++ b/pkgs/google-api-core-grpc.txt
@@ -1,0 +1,9 @@
+win-64/google-api-core-grpc-1.20.0-hc8dfbb8_0.tar.bz2
+win-64/google-api-core-grpc-1.20.0-h32f6830_0.tar.bz2
+win-64/google-api-core-grpc-1.20.0-h9f0ad1d_0.tar.bz2
+osx-64/google-api-core-grpc-1.20.0-h32f6830_0.tar.bz2
+osx-64/google-api-core-grpc-1.20.0-hc8dfbb8_0.tar.bz2
+osx-64/google-api-core-grpc-1.20.0-h9f0ad1d_0.tar.bz2
+linux-64/google-api-core-grpc-1.20.0-h9f0ad1d_0.tar.bz2
+linux-64/google-api-core-grpc-1.20.0-h32f6830_0.tar.bz2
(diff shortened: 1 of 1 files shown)


beckermr:
This should be fixed via a patch on the repo data.

s22chan:
@beckermr can you elaborate? is there a procedure to hack the repo data? the build for the google-api-core feedstock shouldn't even pass if grpc is a mandatory split package that needs to be published.

beckermr:
Yes. We have the conda-forge-repodata-patches feedstock.

s22chan:
https://github.com/conda-forge/conda-forge-repodata-patches-feedstock, https://github.com/conda-forge/conda-forge-repodata-patches-feedstock/tree/master/recipe, and https://anaconda.org/conda-forge/conda-forge-repodata-patches don't have any instructions on how you would do the patch so I'm afraid I still don't follow you on how this can be achieved.

isuruf:
You'll have to read the code in https://github.com/conda-forge/conda-forge-repodata-patches-feedstock/blob/master/recipe/gen_patch_json.py

s22chan:
I'm not sure if we should leave the package in there (untested), but I didn't see the errors in https://github.com/conda-forge/google-api-core-feedstock/issues/46 and `pip install grpcio==1.29.0` seemed to produce a working environment so the patch is over https://github.com/conda-forge/conda-forge-repodata-patches-feedstock/pull/59

s22chan:
I'm seeing the clobber warnings in https://github.com/conda-forge/google-api-core-feedstock/issues/46 as it seems like `1.20.0` incorrectly packaged urllib3. this package should probably be marked as broken instead of hotpatched

isuruf:
This is a meta package with no files. So I don't see how it can be clobbering any other package.

s22chan:
> This is a meta package with no files. So I don't see how it can be clobbering any other package.

The error in https://github.com/conda-forge/google-api-core-feedstock/issues/46#issuecomment-645031584 implies it's in the `google-api-core` package if I'm reading this right, so technically I guess this change should mark all the subpackages and the main package as broken

s22chan:
https://github.com/conda-forge/cf-mark-broken/pull/86 is probably more correct

PR #89, opened 2020-06-16
Mark builds with unpinned macOS SDK as broken

chrisburr:
Closes https://github.com/conda-forge/root_numpy-feedstock/issues/15
Closes https://github.com/conda-forge/rapidsim-feedstock/issues/7

Checklist:

<!--

      ping @conda-forge/foo
--!>


chrisburr:
They were deployed today and I've already deployed new fixed builds which are otherwise identical. I think it's cleaner to just mark them as broken rather than permanently pollute the repodata patches.

(I'm also fairly sure it's impossible to install them on an older macOS as their dependency on `root` is for a version which is pinned correctly)

isuruf:
@chrisburr, can you send a PR to conda-smithy to add a lint for this?

PR #92, opened 2020-07-01
Mark win-64 scipy packages as broken.

adament:
There is (as far as I know) nothing wrong with these packages in an
isolated sense. However when conda is configured according to the
conda-forge documentation: strict channel priority and higher priority
to conda-forge they lock users into the version of scipy provided by
conda-forge (1 for each python version).

As I understand it building scipy on Windows is a cumbersome and manual
process hence only 1 version of scipy has been built for each python
version. Thus users of conda-forge on Windows are locked into scipy 1.3.1
(1.3.2 for python 3.8), whereas it is my understanding that if these
packages are marked as broken conda will resolve to scipy from pkgs/main
which provides both newer and older versions of scipy.

I am by no means sure if this is the right approach, but I think that while Windows
packages of scipy are not being built regularly and thus locking users into a single
scipy version is non-ideal.

Checklist:

* [x] Added a link to the relevant issue in the PR description. [scipy-feedstock/issues/131](https://github.com/conda-forge/scipy-feedstock/issues/131)
* [ ] Pinged the team for the package.
<!--

      ping @conda-forge/foo
--!>

### DIFF
--- a/pkgs/scipy-win64.txt
+++ b/pkgs/scipy-win64.txt
@@ -1,0 +1,3 @@
+win-64/scipy-1.3.1-py36h29ff71c_0.tar.bz2
+win-64/scipy-1.3.1-py37h29ff71c_0.tar.bz2
+win-64/scipy-1.3.2-py38h582fac2_0.tar.bz2


isuruf:
cc @conda-forge/scipy, @mariusvniekerk

isuruf:
Talked to Marius and other core. They are fine with it.

PR #94, opened 2020-07-02
Mark metric-learn 0.6.0 as broken

bellet:
v0.6.0 had missing requirements after dropping support for Python < 3.6
Users should install v0.6.2

https://github.com/scikit-learn-contrib/metric-learn/pull/299
https://github.com/scikit-learn-contrib/metric-learn/pull/300

@conda-forge/metric-learn 

Checklist:

### DIFF
--- a/pkgs/metric-learn.txt
+++ b/pkgs/metric-learn.txt
@@ -1,0 +1,2 @@
+noarch/metric-learn-0.6.0-py_0.tar.bz2
+noarch/metric-learn-0.6.0-py_1.tar.bz2


beckermr:
Unless the actual package is broken in that it doesn’t work, we need to either patch the metadata or leave the build as is. Conda will prefer the newest version anyways if that is your worry.

bellet:
Actually we had correctly set the requirements in the conda recipe for v0.6.0, unlike the PyPI release. Marking this as broken is mainly for consistency with PyPI, where we marked v0.6.0 as "yanked" (whose effect is similar to "broken" here).

Additionally, noarch/metric-learn-0.6.0-py_0.tar.bz2 has a missing license_file field in the recipe, which was the reason for  noarch/metric-learn-0.6.0-py_1.tar.bz2 

Do you think it is better to just leave these builds as is? Indeed, conda will always install the newest version as the requirements are the same

beckermr:
> Do you think it is better to just leave these builds as is? Indeed, conda will always install the newest version as the requirements are the same

Yup. I think in this case leaving things as-is is completely fine. Closing!

PR #97, opened 2020-07-15
request yamllint 1.24.0 be marked broken

bollwyvl:

Checklist:

on https://github.com/conda-forge/yamllint-feedstock/pull/24, we added some test skips to get the package out, but it turns out it was a real bug, swiftly corrected upstream.  https://github.com/conda-forge/yamllint-feedstock/pull/25 (merged) fixes it. I can verify the `1.24.0` we posted is definitely broken in many locale setups... it's not worth working around, though, and I'd rather not have to support inevitable errors.

ping @conda-forge/yamllint

### DIFF
--- a/broken/yamllint-124.txt
+++ b/broken/yamllint-124.txt
@@ -1,0 +1,1 @@
+noarch/yamllint-1.24.0-pyh9f0ad1d_0.tar.bz2


bollwyvl:
Thanks!

PR #98, opened 2020-07-20
mark all recent curl builds for OS X as broken

wolfv:

Checklist:


ocefpaf:
Closes https://github.com/conda-forge/admin-requests/pull/98

wolfv:
build _2 was using a `*dylib` glob to pull in the libcurl...dylib however, this resulted in a very broken package because the files subsection doesn't operate solely on newly added files but on all files in the prefix.

build _3 is broken for yet unknown reasons (@dopplershift mentioned this on gitter).

isuruf:
@wolfv, can you mark `_3` as not broken?

PR #100, opened 2020-07-22
Mark black-19.10b0-py_2 as broken

tadeu:
because it includes "vendored" packages (such as attr, regex and click)
(it has been fixed in build number 3)

See https://github.com/conda-forge/black-feedstock/issues/13

Checklist:


tadeu:
ping @conda-forge/black

PR #101, opened 2020-07-23
mark noarch/blackd-19.10b0-2.tar.bz2 as broken

carlodri:
it depends on noarch/black-19.10b0-2.tar.bz2 which is itself broken

relevant issue/PR: https://github.com/conda-forge/black-feedstock/issues/13 and  https://github.com/conda-forge/black-feedstock/pull/12

Checklist:

### DIFF
--- a/broken/blackd.txt
+++ b/broken/blackd.txt
@@ -1,0 +1,1 @@
+noarch/blackd-19.10b0-2.tar.bz2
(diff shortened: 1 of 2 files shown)


carlodri:
cc @conda-forge/black

mbargull:
Instead of changing `example.txt` please add a new file.

carlodri:
> Instead of changing `example.txt` please add a new file.

sorry...fixed it now!

mbargull:
Thanks!

PR #110, opened 2020-08-11
Mark latest pip as broken

isuruf:

Checklist:

### DIFF
--- a/broken/pip.txt
+++ b/broken/pip.txt
@@ -1,0 +1,1 @@
+noarch/pip-20.2.2-py_0.tar.bz2


isuruf:
Ref: https://github.com/conda-forge/pip-feedstock/issues/58

cc @conda-forge/pip

CJ-Wright:
any idea why it's breaking?

isuruf:
No idea.

CJ-Wright:
It looks like it isn't even getting the old pip filepath right, since that should be noarch as well

mbargull:
@isuruf, I don't know what's up with it, but I'd assume it's something system-related?
You could rebuild the container image (i.e., doing the `conda`/`pip` updates upfront).

isuruf:
Rebuilding the docker image fixed things. Thanks

isuruf:
I also restarted the dozen or so PRs failing due to this issue.

PR #114, opened 2020-08-22
Mark all Spyder and Spyder-kernels betas as broken

ccordoba12:

Checklist:

## Description

This PR marks as broken all Spyder and Spyder-kernels packages present in the beta channel. They were used for testing purposes during the long cycle that lead to Spyder 4, but they're not necessary anymore.

Besides, now that we started the Spyder 5 cycle (see conda-forge/spyder-feedstock#83 for our first alpha), I think they are interfering with the conda solver because I'm unable to install that version from conda-forge.

Ping @conda-forge/spyder
Ping @conda-forge/spyder-kernels


jjhelmus:
**Note that strict channel priority may have removed packages required for satisfiability.**
Strict channel priority is the issue here.  If the beta label does not provide all the required packages then strict channel priority should be disabled.

ccordoba12:
> If the beta label does not provide all the required packages then strict channel priority should be disabled.

You're totally right @jjhelmus! Thanks a lot for your help!

I had set the `strict` `channel_priority` setting, which was preventing me from picking up some packages from the `beta` label and others from the `main` one. After restoring it to `flexible`, things worked as expected.

Sorry for wasting your time with this.

PR #118, opened 2020-08-27
Remove incorrectly named `python-gssapi` package

dhirschfeld:

Checklist:

### DIFF
--- a/broken/python-gssapi.txt
+++ b/broken/python-gssapi.txt
@@ -1,0 +1,9 @@
+win-64/python-gssapi-1.6.9-py38hf317e7c_0.tar.bz2
+win-64/python-gssapi-1.6.9-py36hf9ada80_0.tar.bz2
+win-64/python-gssapi-1.6.9-py37h12bf67a_0.tar.bz2
+osx-64/python-gssapi-1.6.9-py37hecba344_0.tar.bz2
+osx-64/python-gssapi-1.6.9-py38hd1f13c4_0.tar.bz2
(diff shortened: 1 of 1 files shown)


beckermr:
Please make sure to address the checklist above.

dhirschfeld:
This package was added with the incorrect name. To avoid confusion with the conda-forge package not being the same as the pypi package it has been renamed in https://github.com/conda-forge/python-gssapi-feedstock/pull/1

dhirschfeld:
Since the package is less than an hour old I'm pretty sure the only usages are from the related packages I've got on staged-recipes:

* https://github.com/conda-forge/staged-recipes/pull/12469
* https://github.com/conda-forge/staged-recipes/pull/12475

That being the case, I'd be happy for these packages to just be deleted.

beckermr:
we don't delete things, but I am happy to merge this. the tests need to pass first though

isuruf:
I'm not sure if `gssapi` is the best name here as it doesn't indicate that it's python. Is there a C/C++ package named `gssapi`?

dhirschfeld:
I don't think there is a C/C++ project. To build the package it used the `krb5` package which provided the headers/libs for gssapi

dhirschfeld:
The deprecated `python-gssapi` pypi package has the last version as 0.6.4 whereas the version of this package is 1.6.9. Since the old package will never publish another release there will never be any clashes so we could jsut leave the name as-is - thoughts?

dhirschfeld:
I think I'm now inclined to leave it as is so I'll go ahead and close this. If there are any objections it can always be re-opened.

PR #123, opened 2020-09-14
pulp[version=2.3,build_number=0] broken

mbargull:
Newer PuLP versions don't include a default solver.
`pulp[version=2.3,build_number=1]` has a dep on external LP solvers now.

cc @johanneskoester @conda-forge/pulp


mbargull:
Thanks, as always!

PR #131, opened 2020-10-01
Update example.txt

berkelem:
Label BUSCO v4.1.3 as broken

There was a bug in the source code which has been fixed for v4.1.4.

Checklist:

### DIFF
--- a/broken/broken.txt
+++ b/broken/broken.txt
@@ -1,0 +1,1 @@
+noarch/busco-4.1.3-py_0.tar.bz2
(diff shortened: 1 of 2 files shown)


isuruf:
Please don't edit the example file.

berkelem:
> A source code bug is not a good enough reason to mark a package as broken.

How else can I prevent users from using this version?

beckermr:
You cannot. Once a package is shipped, it is always shipped. Conda will install the latest version in new envs or when envs are updated. Marking a package here won't change existing installs of it.

beckermr:
@berkelem I am going to close this issue. Please feel free to reopen it if you feel differently or want to chat more!

PR #141, opened 2020-10-13
Mark pytest noarch build as broken

isuruf:

Checklist:

ref: https://github.com/conda-forge/pytest-feedstock/issues/127

cc @conda-forge/pytest

### DIFF
--- a/broken/pytest.txt
+++ b/broken/pytest.txt
@@ -1,0 +1,1 @@
+noarch/pytest-3.2.2-py_0.tar.bz2


isuruf:
@conda-forge/bot, this might make some recipes not solvable until pytest is migrated, but it's the correct thing to do

beckermr:
automerge is not running here - also is github actions queued?

isuruf:
https://www.githubstatus.com/incidents/nq2s4gt7xs1w

PR #142, opened 2020-10-14
remove redis

ocefpaf:
This is a duplicate of redis-py and not the same package as redis from defaults. We should remove this ASAP.


beckermr:
Did you archive the feedstock?

beckermr:
Looks like yes.

PR #145, opened 2020-10-18
remove bad pinned proj cartopy

ocefpaf:
@xylar do you mind taking a quick look? If I'm correct the affected ones are build numbers 0 and 1 for 0.18 only.


isuruf:
cc @conda-forge/cartopy 

I think I'm missing the context here. Any issue or PR that I can look at?

xylar:
These builds did not constrain `proj`, and as a result we're getting installs that can use an older, incompatible proj.  Upstream didn't specify that `proj` is a dependency (thus, I removed it as a dependency), but @ocefpaf noticed that this was causing trouble and re-introduced it in build 2.  The same may also be true of `geos`.

Here are the 2 PRs for the bad builds:
https://github.com/conda-forge/cartopy-feedstock/pull/88
https://github.com/conda-forge/cartopy-feedstock/pull/89
Note that proj gets removed in https://github.com/conda-forge/cartopy-feedstock/pull/88 and re-added in https://github.com/conda-forge/cartopy-feedstock/pull/91

Here is an issue where we discuss the issue:
https://github.com/conda-forge/cartopy-feedstock/issues/93

xylar:
> @xylar do you mind taking a quick look? If I'm correct the affected ones are build numbers 0 and 1 for 0.18 only.

@ocefpaf, that's correct.  If `geos` is also a problem, though, all builds of 0.18 will be affected and only by new PR will fix that: https://github.com/conda-forge/cartopy-feedstock/pull/94

xylar:
@isuruf and @ocefpaf, I'm seeing these packages labeled as broken but still in `main`:
https://anaconda.org/conda-forge/cartopy/files
Is this expected?

I am still able to install build 0:
```
conda create -y -n test python=3.8 cartopy=0.18.0 proj=6.3.1
```
...
```
  cartopy            conda-forge/linux-64::cartopy-0.18.0-py38h172510d_0
```
It could be that everything is fine and I have a local cache that I can't delete because it's already being used in another env.

xhochy:
Yes, that is expected. We remove the `main` label on the CDN via repodata patching. If you look at the plain `repodata.json` returned by the CDN Url, the package shouldn't be listed anymore.

xylar:
Thanks @xhochy.

ocefpaf:
Thanks @xylar for taking care of this.
Sorry @isuruf, I sent the PR and did not add the proper context.
Also thanks @xhochy for the `main` explanation.

@xylar I'll look into the `geos` issue today. It is probably broken too but maybe we are in the "clear" b/c other deps, like shapely, are bringing the right one.

PR #147, opened 2020-10-19
remove aiohttp bn 0

ocefpaf:
This build number has bad dependencies and are causing trouble in downstream packages. Ideally we should repopatch it but bn 1 is fixes this and I'm not sure it is worth doing the patch.


beckermr:
I'm ok with this since the package is pretty new. @conda-forge/core?

isuruf:
cc @conda-forge/aiohttp

PR #152, opened 2020-10-28
Readd `cudatoolkit` version `11.0.3`

jakirkham:
These were marked as `broken` to avoid causing issues earlier on. We've now relabeled them as `test`, but we can't access them for testing atm because they are still marked as `broken`. So this should fix that.

https://github.com/conda-forge/cudatoolkit-feedstock/issues/15#issuecomment-718143922

<hr>

Checklist:

cc @conda-forge/cudatoolkit @scopatz @isuruf

### DIFF
--- a/not_broken/cudatoolkit11.txt
+++ b/not_broken/cudatoolkit11.txt
@@ -1,0 +1,2 @@
+linux-64/cudatoolkit-11.0.3-he1b5a44_0.tar.bz2
+win-64/cudatoolkit-11.0.3-ha925a31_0.tar.bz2


scopatz:
Looks fine to me!

isuruf:
> We've now relabeled them as test, but we can't access them for testing atm because they are still marked as broken

What does this mean?

> So this should fix that.

This would fix nothing

jakirkham:
> > We've now relabeled them as test, but we can't access them for testing atm because they are still marked as broken
> 
> What does this mean?

https://github.com/conda-forge/cudatoolkit-feedstock/issues/15#issuecomment-718143922

> > So this should fix that.
> 
> This would fix nothing

Not sure I follow. We marked them `broken` before ( https://github.com/conda-forge/admin-requests/pull/143 ). We need to revert that somehow. Relabeling on Anaconda.org is not sufficient to revert that (cc @beckermr to confirm/deny).

beckermr:
So marking them broken is applying a label. If you use this repo to mark them unbroken it will remove that label. However, if you only want to leave the package in the `test` channel and not on `main`, then there is no reason to use an admin request which also rebuilds the repodata patches.

jakirkham:
It's possible I misread things (has been a busy morning), but I got the impression people were having trouble installing `cudatoolkit` version `11.0.3`. So was trying to fix that. If I misunderstood or that is not needed, happy to close. If that is still needed and this is not the right solution, then would appreciate some guidance in figuring out what the right fix is.

beckermr:
which channels do you want the package on?

jakirkham:
Eventually we will want to move all of them to `main`. Right now we just have them in the `test` label until we confirm they are working for folks and before we "release" them.

beckermr:
OK. So you want them to be marked as main, broken and test. This will sync them to the CDN, leave them off the main channel, and have them on the test channel.

isuruf:
> This will sync them to the CDN, leave them off the main channel, and have them on the test channel.

That's really not necessary. labels like `test` are not served through the CDN.

beckermr:
Yup that's right.

jakirkham:
Alright asked people to confirm this is working ok and it seems it is ( https://github.com/conda-forge/cudatoolkit-feedstock/issues/15#issuecomment-718234840 ). Sorry for the noise

PR #159, opened 2020-11-04
Mark as broken noarch/grpcio-health-checking-1.33.2-py*_0.tar.bz2 (bad protobuf pin)

bollwyvl:

Checklist:

ping @conda-forge/grpcio-health-checking (it's just me)

@s22chan brought https://github.com/conda-forge/grpcio-health-checking-feedstock/issues/1 to my attention (thanks!): it's the first release of the package, and has <50 downloads. The upstream metadata, while technically accurate, doesn't reflect that the bulk of the code is generated with a _signficantly_ newer `protoc` than documented. 

https://github.com/conda-forge/grpcio-health-checking-feedstock/pull/2 added a lot more robustness to the testing, and should prevent this _particular_ problem from occurring in the future, though I keep finding more to not like about packaging protobuf-derived things...

### DIFF
--- a/broken/grpcio-health-checking.txt
+++ b/broken/grpcio-health-checking.txt
@@ -1,0 +1,1 @@
+noarch/grpcio-health-checking-1.33.2-pyhd8ed1ab_0.tar.bz2


bollwyvl:
Thanks!

PR #160, opened 2020-11-05
Mark mpifx-20.1 versions as broken

awvwgk:
- upstream switched from year based versioning (20.x) to semantic versioning 1.0
- only uploaded this one version so far, probably easier to mark as broken as to introduce a version epoch
- no dependents in regro/libcfgraph found
- https://github.com/conda-forge/mpifx-feedstock/pull/3 creates the new 1.0 version

---

Checklist:

* [x] Pinged the team for the package for their input. @conda-forge/mpifx

### DIFF
--- a/broken/mpifx.txt
+++ b/broken/mpifx.txt
@@ -1,0 +1,12 @@
+osx-64/mpifx-20.1-hf924dc7_1.tar.bz2
+osx-64/mpifx-20.1-h4c69643_1.tar.bz2
+osx-64/mpifx-20.1-h0ee35d8_1.tar.bz2
+osx-64/mpifx-20.1-h798856c_1.tar.bz2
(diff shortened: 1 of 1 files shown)


beckermr:
I'm ok with this. Semver is preferred imho.

beckermr (approved):

isuruf:
This is what epochs are for. For existing environments, marking them as broken will not upgrade them, because it's a downgrade

beckermr:
We talked about this on gitter and I suggested epochs. My impression here was that the package was not used enough to warrant one.

beckermr:
@isuruf I should add that I am happy to revisit this decision, but the package is only 3 months old and has only 1.5k downloads so something breaking seemed ok.

PR #161, opened 2020-11-09
Remove pdpipe v0.0.521

Silun:

Checklist:


Silun:
This package was the initial commit of a recipe that had some packaging issues. To test a fix, I forked the repo and regrettably forgot to add a "." in the version number, thus bumping it way too high. The recipe was added before I could PR a fix to the original repo and now this broken package is in the channel. I am trying to remove it so the real packages with correct versions from the original repo can be used.

PR #165, opened 2020-11-18
Mark old r-magrittr 2.0.1 packages as broken.

cbrueffer:
These packages should not have been noarch.

Checklist:

### DIFF
--- a/broken/r-magrittr.txt
+++ b/broken/r-magrittr.txt
@@ -1,0 +1,2 @@
+noarch/r-magrittr-2.0.1-r36hc72bb7e_0.tar.bz2
+noarch/r-magrittr-2.0.1-r40hc72bb7e_0.tar.bz2


ocefpaf:
The GH action used here is broken :-(
We can merge the PR but the removal probably won't work until we fix the action or change to another one.

beckermr:
What is broken?

beckermr:
Rebase this pr on master. Should be fixed.

ocefpaf:
The conda GHA sets the environmental variable with a deprecated syntax. I think people are fixing it upstream, maybe there is a new version. I'll check it later but I'm not too familiar with that Action. I use `s-weigand/setup-conda` instead b/c it uses the pre-installed niniconda.

beckermr:
Yep I change the setup mini conda to v2 which fixes this. Rekick the ci maybe?

beckermr:
I'm gonna try merging. If master passes we are fine.

beckermr:
Looks like it worked!

PR #166, opened 2020-11-20
Mark openexr 2.5.0 as broken

hmaarrfk:
There was an so name change between 2.5.0 and 2.5.1

https://github.com/conda-forge/conda-forge-pinning-feedstock/pulls?q=is%3Apr+is%3Aclosed

We want to pin to 2.X, but this version needs to be omitted.


isuruf:
this needs to be a separate file

hmaarrfk:
thanks

PR #211, opened 2021-01-08
Mark msmpi as broken

leofang:

Checklist:

As a follow-up of conda-forge/msmpi-feedstock#1 and requested by @isuruf, the current `msmpi` package (build 0) should be marked as broken due to its missing dependence on `mpi`.

cc: @conda-forge/msmpi @mccarthyryanc

### DIFF
--- a/broken/msmpi.txt
+++ b/broken/msmpi.txt
@@ -1,0 +1,1 @@
+win-64/msmpi-10.1.1-h2845e6d_0.tar.bz2


leofang:
Thanks, @beckermr!

@mccarthyryanc In case you don't know already, the effect of this PR is:
- Attach a broken label on the file (see the red circle)
EED33781-8C65-4B09-96FF-F10A0F36B421
- Make it unavailable for normal install (but can still be forced to install as it's just archived, not deleted)

PR #227, opened 2021-02-02
Mark *lal*-7.1.1 builds 0-2 as broken

duncanmmacleod:
Please mark *lal*-7.1.1 builds 0-2 as broken because they didn't include a necessary `run_constrained` tag, see conda-forge/lal-feedstock#50. I am the maintained of that feedstock and apologise for having borked the 7.1.1 update in the first place.

Checklist:


beckermr:
This can be fixed with a repodata patch. Why is that not the preferred solution here?

duncanmmacleod:
> This can be fixed with a repodata patch. Why is that not the preferred solution here?

@beckermr, because I don't really know what that is and don't know how to prepare one, please advise.

chrisburr:
@duncanmmacleod It's a way conda-forge can change the metadata of existing packages (it's preferred to minimise breakage in existing environments).

It's done modifying `recipe/gen_patch_json.py` in `conda-forge-repodata-patches-feedstock`. You can find an example of one I wrote recently for pip at: https://github.com/conda-forge/conda-forge-repodata-patches-feedstock/pull/116/ In this case you probably want to look for the version and `record['build'].endswith(("_0", "_1", "_2", "_100", "_101", "102"))`.

beckermr:
thanks for explaining @chrisburr!

PR #232, opened 2021-02-16
Mark ipykernel 5.0.0 noarch as broken

ericpre:
Similar as #231, but for ipykernel 5.0.0 noarch packages - obviously, conda is now picking up [one of the these two](https://anaconda.org/conda-forge/ipykernel/files?version=5.0.0)!
From looking at https://conda.anaconda.org/conda-forge/noarch/repodata.json.bz2, there shouldn't be any other!

Checklist:

ping @conda-forge/ipykernel

### DIFF
--- a/broken/ipykernel_5.0.0_noarch.txt
+++ b/broken/ipykernel_5.0.0_noarch.txt
@@ -1,0 +1,3 @@
+noarch/ipykernel-5.0.0-py_0.tar.bz2
+noarch/ipykernel-5.0.0-pyh24bf2e0_1.tar.bz2
+


beckermr:
@conda-forge/core we should find a better way here

beckermr:
I merged a previous one of these but we should not do more and we might want to backout the old one.

isuruf:
noarch packages were a mistake. so, it's correct to mark these as broken

isuruf (approved):

beckermr:
Thanks @isuruf!

PR #212, opened 2021-01-08
Mark Rust 1.49 as broken on macOS x86_64

pkgw:
Cf conda-forge/rust-feedstock#73. It looks like the latest version of Rust accidentally started using a symbol that's not available on our target version of 10.9.

* [x] Make sure your package is in the right spot (`broken/*` for adding the
  `broken` label, `not_broken/*` for removing the `broken` label)
* [x] Added a description of the problem with the package in the PR description.
* [x] Added links to any relevant issues/PRs in the PR description.
* [x] Pinged the team for the package for their input.

ping @conda-forge/rust

PR #236, opened 2021-02-23
Not broken: osx-64/rust-1.49.0-h7e1429e_0.tar.bz2

mbargull:
This has been previously marked broken in https://github.com/conda-forge/admin-requests/pull/212 due to https://github.com/conda-forge/rust-feedstock/issues/73 .

(Barring a potentially missing `__osx >=10.10` dep, which is minor concern,)
The package itself should be fine/not broken, but we needed to ensure
it's not being used with our current deployment target, which was done in
https://github.com/conda-forge/conda-forge-pinning-feedstock/pull/1233

Checklist:

ping @conda-forge/rust, @conda-forge/rust-activation


mbargull:
ATTN: I changed the logic of the PR check here. For `not_broken/*` this now checks against `conda-forge/label/broken`.

beckermr:
@pkgw @conda-forge/rust ?

PR #237, opened 2021-02-27
mark noarch/spacy-model-en_core_web_*_0 broken

bollwyvl:

Checklist:

Notes:
- reported on https://github.com/conda-forge/spacy-model-en_core_web-feedstock/issues/11
- affecting upstream https://github.com/conda-forge/allennlp-semparse-feedstock/pull/2
  - due to some weird pins, this was pulling `spacy=2` and `spacy-model-en_core_web_sm =3`
  - this hadn't really appeared before, as the models "pretty much worked" in `spacy 2`
- replacements available from https://github.com/conda-forge/spacy-model-en_core_web-feedstock/pull/12
  - now uses `run_constrained`, and at least points to the official compatibility matrix
    - in the future, might try to figure out a way to validate them...

ping @conda-forge/spacy-model-en_core_web @h-vetinari

### DIFF
--- a/broken/spacy-model-en_core_web-3.0.0_0.txt
+++ b/broken/spacy-model-en_core_web-3.0.0_0.txt
@@ -1,0 +1,3 @@
+noarch/spacy-model-en_core_web_sm-3.0.0-pyh44b312d_0.tar.bz2
+noarch/spacy-model-en_core_web_md-3.0.0-pyh44b312d_0.tar.bz2
+noarch/spacy-model-en_core_web_lg-3.0.0-pyh44b312d_0.tar.bz2


wolfv:
Are other languages fine (if they exist on conda-forge)?

bollwyvl:
Dunno... I don't build them!

bollwyvl:
@wolfv to your question: it looks like basically none of the other models are particularly maintained on conda-forge.

It _appears_ that all of the models are _tagged_ from the same upstream (not surprising, as they are templated), and _very probably_ uploaded within a window where the bot could find all together.

For this kinda thing, it would likely make sense to just have one mega-feedstock, likely just have it be long and verbose. I'll look into pulling all of those into `en`.

bollwyvl:
Created:

https://github.com/conda-forge/spacy-model-en_core_web-feedstock/issues/13

I'm happy to do the initial work (probably not today, though), and press the occasional button, but would welcome any help.

PR #239, opened 2021-03-08
mark noarch/nteract-scrapbook-0.5.0-*_1 broken

bollwyvl:

Checklist:

The upstream package changed the distribution name (dropping the `nteract-` prefix). I added an alias package for depending on the old name, but it's causing issues for downstreams.

- added in https://github.com/conda-forge/nteract-scrapbook-feedstock/pull/6
- downstream issue reported on https://github.com/conda-forge/nteract-scrapbook-feedstock/issues/7
- fixed with https://github.com/conda-forge/nteract-scrapbook-feedstock/pull/8

@conda-forge/nteract-scrapbook

### DIFF
--- a/broken/nteract-scrapbook-050.txt
+++ b/broken/nteract-scrapbook-050.txt
@@ -1,0 +1,1 @@
+noarch/nteract-scrapbook-0.5.0-pyh44b312d_1.tar.bz2


beckermr:
Removing a package due to a pip check failure is not really needed. 

If pip check fails, that's fine. Conda is not pip and the metadata could simply be different.

beckermr (changes requested):
This pr is not needed.

beckermr:
The solution here is to remove pip check from the recipe of the offending feedstock. Or you can write some code to ignore certain errors.

bollwyvl:
The problem is I (as @conda-forge/nteract-scrapbook) created a package which does _not_ exist on pip (`nteract-scrapbook 0.5.0`) in the thinking that it would be _easier_ to keep working in the future, but this is not the case. In this case, we're dealing with `airflow`-related stuff, which is all but impossible to test in any thorough way, and keep on top of the huge list of dependencies, which `pip check` at least lets us do.

I concur that having a conda-forge-specific `pip check` (which _didn't require pip_!) would be more useful in the long run... except that this metadata actually _does_ get used at runtime by certain parts of the python (not pip) system, namely `entry_points`, which actually check that all of this stuff is consistent. We've encountered this on `python-language-server`, `pytest`, etc. which make significant use of that packaging feature.

beckermr:
I don't personally find this argument convincing. I'm happy to keep chatting though.

cc @conda-forge/core for other opinions

We should be able to create a simple feedstock that holds a python script that runs pip check and then ignores some errors selectively. This tool could then be used by a lot of recipes. This is a much better solution than pulling a package simply because it got renamed on conda.

PR #246, opened 2021-03-29
remove libdap4 builds 0 and 1 for the 3.20.6 version

ocefpaf:
@isuruf these two have a run_exports that does not match the pinning. I guess it makes sense to move them to broken, right?


isuruf:
what's the correct `run_exports`?

ocefpaf:
> what's the correct `run_exports`?

We went with x.x.x, in https://github.com/conda-forge/libdap4-feedstock/pull/40/files

PR #252, opened 2021-04-06
mark all openmm-torch with build number <=2 as broken

jaimergp:
OpenMM-Torch is an OpenMM plugin that links to the C++ API of PyTorch. Initially released for PyTorch 1.7, the  `pin_compatible` arguments were too flexible, and PyTorch does not guarantee ABI stability within releases, as per https://github.com/conda-forge/pytorch-cpu-feedstock/issues/40#issuecomment-803133327

As a result, OpenMM-Torch (build numbers <= 2) will happily install alongside PyTorch 1.8. This has been [addressed](https://github.com/conda-forge/openmm-torch-feedstock/pull/5) in build numbers >=3, but I think the previous attempts should be marked as broken. This could be handled in repodata patches as well, but given the minor impact of this package, I don't think it's necessary to go through that process.

Also, build number 2 was a result of an accidental push to a repo branch (instead of a fork).

Checklist:

  for token resets)


jaimergp:
Thanks @beckermr!

PR #253, opened 2021-04-07
Mark jupyter_client=6.1.13 as broken

davidbrochart:

Checklist:

  for token resets)

Need to pin python >= 3.6
Related issue: https://github.com/jupyter/jupyter_client/issues/634

### DIFF
--- a/broken/jupyter_client.txt
+++ b/broken/jupyter_client.txt
@@ -1,0 +1,1 @@
+noarch/jupyter_client-6.1.13-pyhd8ed1ab_0.tar.bz2


beckermr (changes requested):
This package seems prominent enough that a repodata patch would be better.

davidbrochart:
But then we would have conda and PyPI out of sync, since we can't patch on PyPI?

beckermr:
Ah you removed the pypi release?

davidbrochart:
Not yet, but we have no choice but to yank the PyPI release.

beckermr:
OK. yanking releases is a last resort but sometimes needed. LGTM.

PR #259, opened 2021-04-28
Mark `librmm` build number `0` as broken

jakirkham:
These accidentally included some `spdlog` components that shouldn't have been in there. We've since fixed the issue ( https://github.com/conda-forge/librmm-feedstock/pull/3 ). Though it would be good to remove these packages so that users environments don't accidentally get clobbered by them

cc @isuruf @kkraus14 @leofang


jakirkham:
Thanks Matt! 😄

PR #275, opened 2021-07-21
move all non-shared-builds of lemon to broken

k-dominik:
All builds prior to the `_2` (linux, osx), and `_3` (win) were static builds in lemon. This means they were useful in context of building other packages. Currently [the solver seems to pick up](https://github.com/conda-forge/lemon-feedstock/issues/23) the non-shared versions of this library when installing vigra (afaik the only downstream dependency of lemon).

So - I don't think there will be any negative side effects to removing the non-shared libs here.

I'm not sure if this qualifies to adding the `broken` label. (I probably would remove main and add `static_lib` or something like this).

Checklist:

  for token resets)

ping @conda-forge/lemon

### DIFF
--- a/broken/lemon-non-shared.txt
+++ b/broken/lemon-non-shared.txt
@@ -1,0 +1,16 @@
+linux-64/lemon-1.3.1-0.tar.bz2
+linux-64/lemon-1.3.1-1.tar.bz2
+linux-64/lemon-1.3.1-h338f22d_1.tar.bz2
+osx-64/lemon-1.3.1-0.tar.bz2
+osx-64/lemon-1.3.1-1.tar.bz2
+osx-64/lemon-1.3.1-h338f22d_1.tar.bz2
+win-32/lemon-1.3.1-vc14_0.tar.bz2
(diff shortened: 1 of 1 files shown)


beckermr:
I think this is ok. any thoughts @conda-forge/core?

k-dominik:
Hi @beckermr, thx for merging. Is it expected that the main label still stays on these packages? For me conda will still pick those up and result in broken envs.

beckermr:
Yes this is normal. The main label will remain but the packages won't be picked up. We use a special mechanism in anaconda.org repodata patches that removes the packages from the final package index. 

You'll have to wait about 1-2 hours for everything to resync. At that point, the packages won't be available.

k-dominik:
> Yes this is normal. The main label will remain but the packages won't be picked up. We use a special mechanism in anaconda.org repodata patches that removes the packages from the final package index.
> 
> You'll have to wait about 1-2 hours for everything to resync. At that point, the packages won't be available.

Thank you @beckermr ! It all works nicely now :)

stuarteberg:
Thanks, @k-dominik and @beckermr !

PR #295, opened 2021-08-29
packages are broken for various reasons

MarDiehl:
When restructuring the damask package (split into 3 packages + metapackage, enabled support for MPICH, enabled osx and win support) I tried out many things and created a bunch of broken packages.
Some of them might technically work, but since all of them have been created in the last few days, it should not cause too much confusion to simply mark all but the second but last as broken.

Problems include

- metapackage empty due to unresolvable dependencies
- inconsistencies among different OS
- weird/superflouous dependencies

I don't have an issue (yet) on GitHub, but the topic has been discussed on Gitter today
@navyRUB 
@jan-janssen 

Checklist:

  for token resets)


beckermr:
Please do not delete the example.

beckermr (changes requested):
See above.

MarDiehl:
should keep example.txt

PR #298, opened 2021-09-03
Remove alpha packages for jupyter-sysml-kernel

ivan-gomes:
Kindly requesting removal of alpha (only used for internally testing; has not been distributed/publicized) packages of jupyter-sysml-kernel - ref: https://github.com/conda-forge/jupyter-sysml-kernel-feedstock. These alpha packages were unfortunately published with calendar versioning, e.g. `2021.05a`. This has been corrected with the first non-alpha being published as `0.14.1`, but because of semver rules `2021.05a` still shows up as the latest and default version for the package which is problematic for installing users.

Checklist:

  for token resets)

ping @conda-forge/jupyter-sysml-kernel 

P.S. I am the primary maintainer of this recipe.

### DIFF
--- a/broken/jupyter-sysml-kernel.txt
+++ b/broken/jupyter-sysml-kernel.txt
@@ -1,0 +1,7 @@
+noarch/jupyter-sysml-kernel-2021.05a-pyhd8ed1ab_1.tar.bz2
+noarch/jupyter-sysml-kernel-2021.05a-pyhd8ed1ab_0.tar.bz2
+noarch/jupyter-sysml-kernel-2021.05-pyhd8ed1ab_0.tar.bz2
+noarch/jupyter-sysml-kernel-2021.03a-pyhd8ed1ab_0.tar.bz2
+noarch/jupyter-sysml-kernel-2021.04-pyhd8ed1ab_0.tar.bz2
+noarch/jupyter-sysml-kernel-2021.03-pyhd8ed1ab_0.tar.bz2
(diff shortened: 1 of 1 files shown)


isuruf:
I don't understand. Latest release https://github.com/Systems-Modeling/SysML-v2-Pilot-Implementation/releases seems to be calver.

ivan-gomes:
> I don't understand. Latest release https://github.com/Systems-Modeling/SysML-v2-Pilot-Implementation/releases seems to be calver.

Thanks for looking into it @isuruf, much appreciated! We do use calver for Git tags, but all of our packages/artifacts are semver, ex. https://github.com/Systems-Modeling/SysML-v2-Pilot-Implementation/packages/597995 and https://www.npmjs.com/package/@systems-modeling/jupyterlab-sysml. It seems that semver has the least friction across package/artifact repositories, so we're standardizing on it.

ivan-gomes:
Ping @conda-forge/core and thanks in advance. Don't mean to bug, but we are hoping to have this resolved so it can be included in public communication.

ivan-gomes:
@conda-forge/core, could you kindly help us with this? Our next release window is this coming week and resolution of this issue is necessary to include conda-forge in it. 

@isuruf hope you don't mind me tagging you as well.

xhochy:
Thanks for the link to `npm` that was the only other source to the versioning. You should definitely document the versioning in the main repository.

PR #307, opened 2021-09-15
Mark nbconvert 6.2.0rc0 on main as broken

jasongrout:
We released nbconvert 6.2.0rc0 in the main channel instead of using the rc label by mistake. See https://github.com/conda-forge/nbconvert-feedstock/issues/58#issuecomment-920274460 and the file list at https://anaconda.org/conda-forge/nbconvert/files

Checklist:

  for token resets)

ping @conda-forge/nbconvert

### DIFF
--- a/broken/nbconvert-6.2.0rc0.txt
+++ b/broken/nbconvert-6.2.0rc0.txt
@@ -1,0 +1,17 @@
+win-64/nbconvert-6.2.0rc0-py38haa244fe_0.tar.bz2
+win-64/nbconvert-6.2.0rc0-py37h03978a9_0.tar.bz2
+win-64/nbconvert-6.2.0rc0-py39hcbf5309_0.tar.bz2
+osx-arm64/nbconvert-6.2.0rc0-py38h10201cd_0.tar.bz2
+linux-ppc64le/nbconvert-6.2.0rc0-py37hadc05a3_0.tar.bz2
+osx-64/nbconvert-6.2.0rc0-py37h5186d4c_0.tar.bz2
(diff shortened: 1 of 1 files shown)


broken/nbconvert-6.2.0rc0.txt
```diff
@@ -0,0 +1,17 @@
...
+osx-64/nbconvert-6.2.0rc0-py37hf985489_0.tar.bz2
+osx-64/nbconvert-6.2.0rc0-py39h6e9494a_0.tar.bz2
+osx-64/nbconvert-6.2.0rc0-py38h50d1736_0.tar.bz2
+linux-ppc64le/nbconvert-6.2.0rc0-py37h35e4cab_0.tar.bz2
+linux-ppc64le/nbconvert-6.2.0rc0-py38hf8b3453_0.tar.bz2
+linux-ppc64le/nbconvert-6.2.0rc0-py39hc1b9086_0.tar.bz2
+linux-64/nbconvert-6.2.0rc0-py39hf3d152e_0.tar.bz2
+linux-64/nbconvert-6.2.0rc0-py38h578d9bd_0.tar.bz2
+linux-64/nbconvert-6.2.0rc0-py37h9c2f6ca_0.tar.bz2
+linux-64/nbconvert-6.2.0rc0-py37h89c1867_0.tar.bz
```
beckermr, line 17:
```suggestion
linux-64/nbconvert-6.2.0rc0-py37h89c1867_0.tar.bz2
```

jasongrout: oops, thanks.

PR #311, opened 2021-09-17
Request token reset for `termcolor-cpp`

adriendelsalle:

Checklist:

  for token resets)

[https://github.com/conda-forge/termcolor-cpp-feedstock](termcolor-cpp) package upload is broken.
See more details on the [CI](https://dev.azure.com/conda-forge/feedstock-builds/_build/results?buildId=377537&view=logs&j=656edd35-690f-5c53-9ba3-09c10d0bea97&t=e5c8ab1d-8ff9-5cae-b332-e15ae582ed2d&l=312)

### DIFF
--- a/token_reset/termcolor-cpp
+++ b/token_reset/termcolor-cpp
@@ -1,0 +1,1 @@
+termcolor-cpp


adriendelsalle:
@beckermr thanks, looks like I missed the `txt` extension and that failed

beckermr:
I just fixed it on master.

PR #317, opened 2021-09-22
remove curl 7.79.0

ocefpaf:
This bug https://github.com/curl/curl/issues/7738 is solved in 7.79.1 and that one will be in the channel soon (https://github.com/conda-forge/curl-feedstock/pull/96) but we need to remove the 7.79.0 to avoid headaches.


isuruf (approved):

isuruf:
cc @conda-forge/curl

PR #338, opened 2021-11-04
Mark CuPy 9.5.0 build 0 packages broken

jakirkham:
A patch was needed to fix CUDA enhanced compatibility support. This was included in build 1 and will be in the next CuPy minor and major releases. To avoid picking up CuPy 9.5.0 packages without this patch, mark build number 0 broken.

xref: https://github.com/conda-forge/cupy-feedstock/pull/153

cc @leofang @conda-forge/cupy

cc @galipremsagar @shwina @ajschmidt8 (for awareness)

<hr>

Checklist:

  for token resets)


galipremsagar (approved):

jakirkham:
Thanks Matt! 😄

PR #339, opened 2021-11-05
Add libcint_0 to broken

BastianZim:

Checklist:

  for token resets)

* The package was accidentally added without `run_exports`, this has been fixed now: https://github.com/conda-forge/libcint-feedstock/commit/39e200ef4525025a1bb6fa60c4061e7a8a086629#diff-e1d8e552330911f9f779f85b6f2c00a15e790dcc3fbb3b28f5da1d660a30c5b8
* I'm the maintainer of the feedstock: https://github.com/conda-forge/libcint-feedstock
* The package was only added to conda-forge yesterday so I think just marking it as broken is fine since no one downloaded it yet probably.


BastianZim:
Thanks!

PR #340, opened 2021-11-06
Mark `pywavelet` Python 3.10 packages broken

jakirkham:
These appeared to pass CI, but actually failed their tests. So mark them broken.

xref https://github.com/conda-forge/pywavelets-feedstock/pull/42#pullrequestreview-799464721

cc @rgommers @conda-forge/pywavelets

Checklist:

  for token resets)


jakirkham:
These were barely uploaded. So may not be available via CDN. Hence the failure above about missing packages

jakirkham:
Converting to draft for the moment. Some more discussion about the packages affected suggests this may be `ppc64le` and not Python 3.10 that is the issue ( https://github.com/conda-forge/pywavelets-feedstock/pull/42#issuecomment-962504719 ). May need to update this to include a different set of packages then

jakirkham:
Superseded by PR ( https://github.com/conda-forge/admin-requests/pull/341 )

PR #349, opened 2021-11-22
Mark `dotnet-sdk` 6.0.100 build 0 as broken

dhirschfeld:

Checklist:

  for token resets)

----

.NET6 added a new sub-folder `sdk-manifests` which wasn't present in .NET5. This new folder wasn't being copied over in the build so the SDK apparently doesn't work for building .NET apps :(

This was brought to our attention and resolved in https://github.com/conda-forge/dotnet-feedstock/pull/45

Since build 0 won't actually work (IIUC) for building .NET apps it should be yanked to avoid others running into the same issue as seen in https://github.com/pythonnet/clr-loader/issues/19


dhirschfeld:
REsolves https://github.com/conda-forge/dotnet-feedstock/issues/46

dhirschfeld:
Ping @conda-forge/dotnet

PR #354, opened 2021-12-02
Mark mesalib with gl builds as broken

hmaarrfk:
These packages were accentally built with `GL.so` which overrides the user's system GL.so file causing all opengl to be software rendered. This is almost certainly not what people want when they are trying to use scientific visualization software that depends on GL.so

See https://github.com/conda-forge/mesalib-feedstock/pull/40#issuecomment-984808054
And an issue raised by a user: https://github.com/conda-forge/mesalib-feedstock/issues/16#issuecomment-984504934

@conda-forge/mesalib do you support marking all 21 builds as broken?

Checklist:

  for token resets)

### DIFF
--- a/broken/mesalib_21_with_gl.txt
+++ b/broken/mesalib_21_with_gl.txt
@@ -1,0 +1,13 @@
+linux-aarch64/mesalib-21.2.5-h0bd06be_2.tar.bz2
+linux-ppc64le/mesalib-21.2.5-ha5911c8_2.tar.bz2
+osx-arm64/mesalib-21.2.5-h25af579_2.tar.bz2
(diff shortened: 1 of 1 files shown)


jakirkham:
Thoughts @conda-forge/mesalib? 🙂

beckermr:
merging since the PR to fix this was merged too.

hmaarrfk:
Thank you!

hmaarrfk:
Was this not supposed to be removed from the main label?

IT seems that the broken packages are still available on main
image


Thanks for the clarification

xhochy:
Broken builds are on `main` and `broken` but patched out of the repodata in the CDN. This prevents the installation of them but the URLs to the packages keep existing, e.g. we don't break lockfiles with the URLs in it.

PR #362, opened 2021-12-17
edgetest-hub broken

fdosani:
Hi @conda-forge/core team. I'd like to mark the edgetest-hub package as broken as we've had to remove it from PyPi and GitHub due to some info being captured in the source code which shouldn't be there. (Happy to explain further via chat or email). I'm hoping to get the releases off the channel until we can refactor/fix it and re-publish it.

Checklist:

  for token resets)


beckermr:
Note the package will still be public but not installable but default. If you need it deleted for some reason, get in touch. We don't do that per our normal policies but we're always happy to chat.

PR #363, opened 2021-12-18
remove freecad packages that uses wrong source-version

looooo:

Checklist:

  for token resets)


beckermr (changes requested):
Please follow the instructions above on notifying teams and linking to issues on feedstocks. If you are the freecad team, then you can simply add that info.

looooo:
@beckermr 
there were some commits backported lately. The mentioned builds do not use the final source. But the builds with build-number `3` should be identical to the older builds (regarding dependencies).
there is no reference issue, sorry.

Not sure what else I should do.

beckermr:
Are the packages actually not functional for some reason?

cc @conda-forge/freecad for viz in case it is not @looooo

looooo:
@beckermr it works, but is missing some backport commits.

beckermr:
Wait are you saying upstream is backporting commits but not releasing new versions?

beckermr:
We don't usually mark stuff broken if it functional and upstream has not pulled it.

beckermr:
Upstream should be making patch releases when backporting and the e should be telling them that.

looooo:
> Wait are you saying upstream is backporting commits but not releasing new versions?

sorry, we did some testing with the builds mentioned here. I know the workflow is not perfect, but adding a `broken` label to these builds is really the simplest way to proceed.

beckermr:
Ok. Please use rc or dev releases on a label in the future.

PR #365, opened 2021-12-21
Mark current orange3 packages for python310 as broken

markotoplak:

Checklist:

  for token resets)

Our team merged conda-forge/orange3-feedstock#107 (Rebuild for python310). Unfortunately, there are problems we did not foresee. To support Python 3.10 `orange3` needs both:
1. PyQt >= 5.15 or corresponding workarounds in code
2. Fixes for deprecated implicit casts to `int` in biolab/orange3

Some discussion about the problem: biolab/orange3#5748

`orange3=3.31.0` thus can not be fixed to support Python 3.10. We thus kindly ask the core team to mark the corresponding packages as broken and we will revert conda-forge/orange3-feedstock#107 for the time being.


markotoplak:
ping @conda-forge/orange3

PR #381, opened 2022-01-26
Remove bad python eccodes batch2

iainrussell:

Checklist:

  for token resets)

This is the second (and should be the last) batch of wrongly versioned python-eccodes packages (packages with version numbers such as "2021.01.0" instead of 'proper' version numbers).


jakirkham:
cc @conda-forge/python-eccodes

PR #391, opened 2022-02-22
Remove uninstallable mpifort

carterbox:
This specific build of mpi-fort is uninstallable because it requires the corresponding compiler build which doesn't exist? Perhaps that build was already removed for being broken, but they forgot to remove this package too?

Checklist:

  for token resets)

### DIFF
--- a/broken/mpi-fort.txt
+++ b/broken/mpi-fort.txt
@@ -1,0 +1,1 @@
+linux-64/openmpi-mpifort-4.0.2-he991be0_3.tar.bz2
--- a/broken/mpi-fort.txt
+++ b/broken/mpi-fort.txt
@@ -1,1 +1,3 @@
 linux-64/openmpi-mpifort-4.0.2-he991be0_3.tar.bz2
+linux-64/openmpi-mpicxx-4.0.2-hc9558a2_3.tar.bz2
+linux-64/openmpi-mpicc-4.0.2-h516909a_3.tar.bz2


carterbox:
Related to https://github.com/conda-forge/staged-recipes/pull/14726

carterbox:
@conda-forge/openmpi

leofang:
Thanks, @carterbox. Good catch. I should have done this in #17... Since you're at it, would you mind also removing `openmpi-mpicc` and `openmpi-mpicxx` with the same version and build numbers? I bet they don't work either.

carterbox:
```       
openmpi-mpicc                  4.0.2      h516909a_3  conda-forge                
openmpi-mpicxx                 4.0.2      hc9558a2_3  conda-forge                
openmpi-mpifort                4.0.2      he991be0_3  conda-forge         
```
These are the package infos. Someone check that I got the tarball names right.

leofang (approved):

carterbox:
There already is a newer build of these packages, but the solver chooses these old ones anyways. I don't know why.

The problem is that these packages have a pinned dependency on a build of openmpi that was already marked as broken, so they cannot be installed.

It seems better to mark these uninstallable packages as broken than to modify every recipe to avoid this specific build of these packages.

jakirkham:
Thoughts @conda-forge/core? 🙂

hmaarrfk:
I believe the issue is that you need to add the cuda compilers so that the recepe generates multiple build strings of mpifort
https://github.com/conda-forge/openmpi-feedstock/blob/master/recipe/meta.yaml#L99

Should look like:
```
        - {{ compiler('c') }}
        - {{ compiler('cxx') }}
        - {{ compiler('fortran') }}
        - {{ compiler('cuda') }}  # [linux64]
```
That way your build matrix gets built out.

hmaarrfk:
Hmm, nevermind. I mean honestly, probably some edge case is getting hit. I'm ok with moving on from this. But hopefully we can write some good guides to compiling with GPU support.

For one, I don't think we've documented good build string conventions and best practicies for pinning gpu/cpu.

jakirkham:
Probably best to continue that discussion in issue ( https://github.com/conda-forge/conda-forge.github.io/issues/1059 ) (assuming this is covering the same thing)

carterbox:
Thanks for merging.

jakirkham:
Nor am I. Though I haven't spent any time digging into this

PR #392, opened 2022-02-24
verilator-feedstock#13 uploaded new packages with same build number

timsnyder:
In conda-forge/verilator-feedstock#13, I screwed up and uploaded packages with the same build number.

not sure if this is a valid reason to mark them broken but I'm starting with this because patching repodata doesn't
seem like it can fix the build number being wrong in the filename.

Checklist:

  for token resets)

### DIFF
--- a/broken/verilator_12.txt
+++ b/broken/verilator_12.txt
@@ -1,0 +1,2 @@
+osx-64/verilator-4.034-hd9580d2_1.tar.bz2
+linux-64/verilator-4.034-h6239696_1.tar.bz2


hmaarrfk:
you could... or maybe you could upload a new build number?

hmaarrfk:
if you want us to delete these, can you confirm that this was the "second" package you uploaded, and not the "first"

timsnyder:
The names in this PR were the second ones uploaded with build number 1 for version 4.034.  I reuploaded them with build number 2 (rebuilt with bumped number).  I circled them in this screenshot 
image

If it isn't going to break anything, then we don't have to delete them.  I just couldn't remember how conda/mamba handles this situation.

hmaarrfk:
Thanks

PR #397, opened 2022-03-03
OpenMM packages for Mac are broken

peastman:
We merged https://github.com/conda-forge/openmm-feedstock/pull/70 two days ago to build Python 3.10 packages for OpenMM.  In addition to building Python 3.10 packages, it also rebuilt the packages for all earlier Python versions.  It appears that all the newly built packages for Mac are broken.  They produce segfaults when anyone tries to use them.  We don't understand why, but the old packages work correctly and the new packages built from the same tag fail.  Presumably this is due to a change to something else in the conda-forge build infrastructure?  See https://github.com/openmm/openmm/issues/3495.

Checklist:

  for token resets)


peastman:
Thanks!

PR #415, opened 2022-03-24
Fix validation error with `omniorb-libs`

jakirkham:
Closes https://github.com/conda-forge/artifact-validation/issues/805
Closes https://github.com/conda-forge/artifact-validation/issues/806
Closes https://github.com/conda-forge/artifact-validation/issues/807

These `omniorb-libs` packages included components of `openssl` that shouldn't have been included. This would result in clobbering `openssl`'s files and potentially issues for packages built on top of them. So here they are marked `broken` to ensure they don't get picked up. The feedstock will need to address this issue further to ensure new packages don't have the same issue.

cc @beckermr @conda-forge/core @conda-forge/omniorb

Checklist:

  for token resets)

### DIFF
--- a/broken/fix_805_validation_omniorb-libs.txt
+++ b/broken/fix_805_validation_omniorb-libs.txt
@@ -1,2 +1,8 @@
+linux-ppc64le/omniorb-libs-4.2.5-h350ef5c_1.tar.bz2
+linux-ppc64le/omniorb-libs-4.2.5-h262a72d_1.tar.bz2
+win-64/omniorb-libs-4.2.5-h8b1b97a_1.tar.bz2
+win-64/omniorb-libs-4.2.5-hc7e8586_1.tar.bz2
 osx-64/omniorb-libs-4.2.5-h942079c_1.tar.bz2
 osx-64/omniorb-libs-4.2.5-hca56917_1.tar.bz2
+linux-64/omniorb-libs-4.2.5-he49606f_1.tar.bz2
+linux-64/omniorb-libs-4.2.5-h727a467_1.tar.bz2
--- a/broken/fix_805_validation_omniorb-libs.txt
+++ b/broken/fix_805_validation_omniorb-libs.txt
@@ -1,0 +1,2 @@
+osx-64/omniorb-libs-4.2.5-h942079c_1.tar.bz2
+osx-64/omniorb-libs-4.2.5-hca56917_1.tar.bz2


broken/fix_805_validation_omniorb-libs.txt
```diff
@@ -0,0 +1,2 @@
+osx-64/omniorb-libs-4.2.5-h942079c_1.tar.bz2
+osx-64/omniorb-libs-4.2.5-hca56917_1.tar.bz2
```
beenje, line 2: Shouldn't windows and linux packages be marked as broken as well?

jakirkham: Do they have the same issue?

beenje: Yes... I noticed the issue on Windows. I checked Linux and it has the same problem, which is consistent.

beckermr: The rest are showing up: https://github.com/conda-forge/artifact-validation/issues/806

Our scanning takes time.

jakirkham: So that shows the Windows packages, but not the Linux packages. Searched the issues there, but only seeing Windows and macOS

beckermr: The bot probably has not gotten to Linux yet maybe.

jakirkham: Ok have included them all. Please recheck 🙂

beenje (approved):

beckermr:
I am 100% sure we should flag the rest of the builds here as broken even if our bots have not found them yet.

beckermr:
There are osx-64 builds missing.

broken/fix_805_validation_omniorb-libs.txt
```diff
@@ -0,0 +1,8 @@
+linux-ppc64le/omniorb-libs-4.2.5-h350ef5c_1.tar.bz2
+linux-ppc64le/omniorb-libs-4.2.5-h262a72d_1.tar.bz2
+win-64/omniorb-libs-4.2.5-h8b1b97a_1.tar.bz2
+win-64/omniorb-libs-4.2.5-hc7e8586_1.tar.bz2
+osx-64/omniorb-libs-4.2.5-h942079c_1.tar.bz2
+osx-64/omniorb-libs-4.2.5-hca56917_1.tar.bz2
```
jakirkham, line 6: Which `osx-64` builds are missing? I only see these two. Please also feel free to push changes to my branch

beckermr:
Oh woops I may have been looking at an old commit! Sorry!

beckermr:
The list looks complete to me. Merging!

jakirkham:
No worries. Also want to make sure we get this right. So appreciate the close review. Thanks Matt! 😄

PR #423, opened 2022-04-07
Remove broken f2c files: build 1002

Areustle:
The build 1002 does not contain the f2c executable and is superseded by build 1004, as merged by https://github.com/conda-forge/f2c-feedstock/pull/13

Checklist:

  for token resets)


jakirkham:
Thanks Alex! 🙏

Should 1003 be removed too?

broken/example.txt
```diff
@@ -1,2 +1,7 @@
 win-64/cf-autotick-bot-test-package-0.4-py38_0.tar.bz2
 win-64/cf-autotick-bot-test-package-0.4-py27_0.tar.bz2
+osx-arm64/f2c-20210323183018-h1c322ee_1002.tar.bz2
+osx-64/f2c-20210323183018-h5eb16cf_1002.tar.bz2
+linux-ppc64le/f2c-20210323183018-hb283c62_1002.tar.bz2
+linux-aarch64/f2c-20210323183018-h4e544f5_1002.tar.bz2
+linux-64/f2c-20210323183018-h166bdaf_1002.tar.bz2
```
jakirkham, line 7: Also these should be added to a new file in the same directory

Areustle:
Build 1003 was not committed as far as I can tell. But if you see an instance of it then yes, remove it.

broken/example.txt
Areustle: You're right. It was late for me when I opened this PR.

jakirkham: No worries. Thanks for fixing that

jakirkham:
Thanks Alex! 😄

PR #431, opened 2022-04-26
Mark compilers 7.5.0 builds that use glibc 2.14 symbols as broken

kkraus14:

Checklist:

  for token resets)

Follow up on https://github.com/conda-forge/ctng-compilers-feedstock/issues/97. The packages listed are supposed to be supported on cos6 environments but got glibc 2.14 symbols into them in the latest builds. Unsure if we want to mark them as broken or use a repodata patch to enforce a glibc constraint instead. There was also some Gitter discussion here: https://gitter.im/conda-forge/conda-forge-compilers?at=6266d2eac61ef0178edc0ac0.

cc @xhochy @isuruf @beckermr @jakirkham for visibility


isuruf:
Repodata patch is preferred.

jakirkham:
How do we want to patch them? Add `__glibc >=2.17` (CentOS 7's GLIBC version) to the packages' `constrains`?

PR #435, opened 2022-05-04
Mark qt-webengine packages for osx-arm64 as broken

Tobias-Fischer:
See https://github.com/conda-forge/qt-webengine-feedstock/pull/5

/cc @hmaarrfk


hmaarrfk:
I unfortunately uploaded empty packages. Sorry.

PR #441, opened 2022-05-24
mark un-extractable package https://conda.anaconda.org/conda-forge/win-64/qt-5.12.9-h556501e_5.tar.bz2

dholth:
This file is corrupt. tar: Truncated input file (needed 81575936 bytes, only 0 available)

In the future it should not be possible to upload packages that can't be extracted


beckermr:
I marked this one by hand. I will kick off a repodata rebuild to propagate it to the channel. Closing!

PR #449, opened 2022-06-05
Mark prematurely-released pymc 4.0.0 build 0 as broken

maresb:
Discussion (and ping) here: https://github.com/conda-forge/pymc-feedstock/issues/40

Checklist:

  for token resets)

### DIFF
--- a/broken/pymc.txt
+++ b/broken/pymc.txt
@@ -1,0 +1,15 @@
+win-64/pymc-4.0.0-py310h5588dad_0.tar.bz2
+win-64/pymc-4.0.0-py37h03978a9_0.tar.bz2
+win-64/pymc-4.0.0-py39hcbf5309_0.tar.bz2
+win-64/pymc-4.0.0-py38haa244fe_0.tar.bz2
+osx-64/pymc-4.0.0-py39h360fd21_0.tar.bz2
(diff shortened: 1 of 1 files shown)


maresb:
I just got :+1: from the active maintainer: https://github.com/conda-forge/pymc-feedstock/issues/40#issuecomment-1147961505

maresb:
Relevant comment copied here:

>Update: we have restructured the feedstock to have two outputs pymc-base and pymc.

> The premature pymc 4.0.0 build 0 does not include mkl-service (or openblas in the case of M1).
>
> On the other hand, build 1 imposes the default of mkl-service (or openblas), but allows for customization via pymc-base.
>
> The way Conda works is that it considers all builds of all versions. Therefore, if one installs pymc >=4.0.0 then the solver could choose either build 0 without mkl-service or build 1 with.
>
> The advantage to marking build 0 as broken is that if a user installs pymc >=4.0.0 then we know that they have BLAS installed.
>
> The disadvantage is mitigated because we now have a properly working build 1 which should transparently "just work" for almost all people who installed build 0 over the previous two days.
>
> Please give feedback, I will need some consensus before I can get https://github.com/conda-forge/admin-requests/pull/449 accepted.

PR #470, opened 2022-08-10
Remove osx arm64 pandoc

ocefpaf:
@msarahan do you mind taking a look if this is OK?

xref.: https://github.com/conda-forge/pandoc-feedstock/pull/116#issuecomment-1210963415


msarahan:
Thanks for taking care of this

PR #475, opened 2022-08-18
Add broken ONNX packages.

janjagusch:
I'm unsure whether this qualifies as broken, but we've released a wrong `onnx` version [here](https://github.com/conda-forge/onnx-feedstock/pull/78). The version should have been `1.12.0`, not `12.0.0`. If you prefer a different solution, please let me know.

Checklist:

  for token resets)


janjagusch:
@xhochy, is this how you'd like to deal with the accidental v12 release?

broken/onnx.txt
```diff
@@ -0,0 +1,15 @@
+linux-ppc64le/onnx-12.0.0-py39he81994a_0.tar.bz2
```
xhochy, line 1: Can you also add `linux-64`, `linux-aarch64` and `osx-arm64` build here, too?

xhochy (changes requested):

broken/onnx.txt
janjagusch: @xhochy, thanks for the thorough review. I have added the missing archs.

xhochy (approved):

PR #476, opened 2022-08-25
Mark broken protobuf packages with incorrect version numbers

xylar:
Because of confusion about protobuf's version numbering scheme
and the way the bot detected it, 2 package versions were introduced
with incorrect version numbers (3.21.1 and 21.2, repsectively).
These should have been 4.21.1 and 4.21.2, respectively.

Relevant discussion is here:
https://github.com/conda-forge/protobuf-feedstock/pull/147
and here:
https://github.com/conda-forge/protobuf-feedstock/pull/146#issuecomment-1149410254

Checklist:

  for token resets)


xylar:
@conda-forge/protobuf, flagging you to make you aware of this.

@tpdownes, also flagging you.  Thank you for your kindness and patience in explaining the problem.  It took me awhile to get it but I'm onboard now and will ask for your approval as I attempt to fix this.

h-vetinari:
> 2 package versions were introduced
> with incorrect version numbers (3.21.1 and 21.2, repsectively).

I would have though 3.21 is alright as it's still following the old naming scheme?



> These should have been 4.21.1 and 4.21.2, respectively.

Afaict, libprotobuf itself should stay at 3.x, and only the python bindings get 4.x

h-vetinari:
> > These should have been 4.21.1 and 4.21.2, respectively.
> 
> Afaict, libprotobuf itself should stay at 3.x, and only the python bindings get 4.x

NVM, this PR just did protobuf anyway, not libprotobuf. We also need to mark libprotobuf 21.x as broken (but not 3.21.1).

xylar:
@h-vetinari, you have that exactly right.  But `libprotobuf` has moved erroneously to 21.x instead of 3.21.x and that need to be fixed as well.

xylar:
@h-vetinari, yes, what you said!

xylar:
Thanks @ocefpaf!!!

xylar:
This helps to assuage some of my guilt at rushing in and messing things up.  Better luck tomorrow...

PR #479, opened 2022-09-02
Mark builds of xeus_octave that do not pin octave as broken

SylvainCorlay:
Cf https://github.com/conda-forge/xeus_octave-feedstock/issues/6

### DIFF
--- a/broken/xeus_octave.txt
+++ b/broken/xeus_octave.txt
@@ -1,0 +1,6 @@
+linux-64/xeus_octave-0.0.1-h2c45cf3_0.tar.bz2
+osx-64/xeus_octave-0.0.1-h124c09a_0.tar.bz2
+linux-64/xeus_octave-0.0.2-h2c45cf3_0.tar.bz2
+osx-64/xeus_octave-0.0.2-h124c09a_0.tar.bz2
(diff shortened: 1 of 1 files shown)


isuruf:
I think, we'd prefer a repodata-patch instead of marking them broken.

SylvainCorlay:
We could do that, although they are also named `xeus_octave` instead of `xeus-octave`. This is merely version 0.0.2 so this should not be too much of an issue.

isuruf:
Quoting from the PR guidelines

> We prefer to patch the repo data (see [here](https://github.com/conda-forge/conda-forge-repodata-patches-feedstock)) instead of marking packages as broken. This alternative workflow makes environments more reproducible.

So, let's follow our own guidelines.

SylvainCorlay:
Another reason why this is really broken is that we did not have [that patch](https://github.com/conda-forge/xeus_octave-feedstock/blob/main/recipe/no-fltk.patch) while conda-forge's octave is build `--without-fltk`.

isuruf:
I'm trying to find a PR for that patch to find an explanation. Could you describe if the package is unusable or it's a warning or something like that?

SylvainCorlay:
It is actually unusable.

isuruf (approved):

PR #481, opened 2022-09-08
Remove chemics package

wigging:

Checklist:

  for token resets)

I would like to remove the chemics package from conda-forge. As the developer of the package, I only want it available via pip and have no need for conda-forge.

### DIFF
--- a/broken/chemics.txt
+++ b/broken/chemics.txt
@@ -1,0 +1,10 @@
+noarch/chemics-22.8-pyhd8ed1ab_0.tar.bz2
+noarch/chemics-21.10-pyhd8ed1ab_0.tar.bz2
+noarch/chemics-21.7-pyhd8ed1ab_0.tar.bz2
+noarch/chemics-21.5-pyhd8ed1ab_0.tar.bz2
+noarch/chemics-21.4-pyhd8ed1ab_0.tar.bz2
+noarch/chemics-20.9-pyh9f0ad1d_0.tar.bz2
+noarch/chemics-20.8-pyh9f0ad1d_0.tar.bz2
+noarch/chemics-20.7-pyh9f0ad1d_0.tar.bz2
+noarch/chemics-20.5-pyh9f0ad1d_1.tar.bz2
+noarch/chemics-20.5-pyh9f0ad1d_0.tar.bz2


wolfv:
@wigging I kinda understand your reasoning, but conda-forge is managed by independent people and your package might also be required by some other packages in the conda-forge ecosystem. Did you receive a lot of bug reports because of the conda-forge package?

Usually the feedstock should not be something "you" have to care about (but rather the person that packages your library for conda-forge).

ocefpaf:
> @wolfv If other packages depend on chemics then it is available via pip.

That thinking excludes all the other users of multiple tools like yum, brew, choco, apt, etc. If a package is OSS and has a proper OSS license, it will have a life outside of the author choice of publication platform. BTW, thanks on a successful and useful project b/c it got the interested of open source users.

> I honestly doubt that it's used by anyone other than myself.

You underestimate the use of your package. Only in [anaconda.org](https://anaconda.org/search?q=chemics) there are two channels that provides it. The conda-forge downloads alone are 9966 today. Also, many others mirror and/or re-package PyPI. I bet there are tons of chemics packages out there in other platforms.

> As the developer of the package, I just don't want to worry about conda installations and just focus on installing with pip.

Fair. You can close any issues folks open upstream about it. In fact, you can even add a template saying that conda is not supported by you. In general folks rarely open issues upstream about downstream packages. And, if you think that there aren't many people using it, there won't be a lot of issues anyway. We hope that (and teach) our users that they must open issues about the package in the feedstock, not upstream.

> The only reason the package is available on conda-forge is because someone else setup the feedstock.

Yes. That is how third party packaging works. In fact two people added it BTW. Note that PyPI is an author publishing platform conda, brew, choco, all linux repositories, conda, etc are not an author publishing platform. They are Open Source Package platforms, and package everything following the rules of the Open Source Licenses in the packages.

wigging:
So if I can't remove chemics from conda-forge, can someone at least remove my GitHub account from conda-forge notifications? Every time I push a new release I get notified to join the conda-forge team on GitHub. I don't want to join the conda-forge team.

ocefpaf:
> So if I can't remove chemics from conda-forge, can someone at least remove my GitHub account from conda-forge notifications? Every time I push a new release I get notified to join the conda-forge team on GitHub. I don't want to join the conda-forge team.

Sure. That is something we can do. We are also very careful to only add folks to agree to it. I'll send a PR to the feedstock removing you ASAP.

Looks like you got added in https://github.com/conda-forge/chemics-feedstock/pull/1

@jan-janssen please be sure to check with maintainers before adding them next time.

PR #486, opened 2022-09-15
Label `arpack` v3.8.0=*_0 and =*_100 as broken

mrakitin:
Per https://github.com/conda-forge/arpack-feedstock/pull/47#issuecomment-1247538671.

Produced with this one-liner:
```bash
$ curl https://anaconda.org/conda-forge/arpack/files 2>&1 | grep '3\.8\.0/download' | awk -F['><'] '{print $3}' | sort -u > arpack.txt
```

---

Checklist:

  for token resets)

### DIFF
--- a/broken/arpack.txt
+++ b/broken/arpack.txt
@@ -1,0 +1,13 @@
+linux-64/arpack-3.8.0-mpi_mpich_h2a544ef_0.tar.bz2
+linux-64/arpack-3.8.0-mpi_openmpi_h5a9222f_0.tar.bz2
+linux-64/arpack-3.8.0-nompi_h0baa96a_100.tar.bz2
+linux-aarch64/arpack-3.8.0-mpi_mpich_hfd77740_0.tar.bz2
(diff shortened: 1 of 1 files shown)


mrakitin:
Thank you for your help, @beckermr Matthew!

Just curious, are the labels added via the GHA workflows in this repo, or is there a separate bot? I don't see the label set in https://github.com/conda-forge/admin-requests/actions/runs/3058630006/jobs/4935056195, maybe it's elsewhere.

mrakitin:
I noticed the `broken` label was added:
image

Should the `main` label still be there?

mrakitin:
I see how it's done with the GHA workflows, cool!
https://github.com/conda-forge/admin-requests/blob/246514993407ca18f33b003db0c343c87cae362a/.github/workflows/master.yml#L59-L64

Still curious about the `main` label.

beckermr:
The main label stays on the but any package with broken is excluded from the metadata

mrakitin:
Thanks, good to know!

PR #488, opened 2022-09-16
mark google-cloud-cpp 2.2.1 as broken

coryan:
The package changes ABI on each patch release (which are rare, but do happen). The 2.2 release only pinned major and minor versions, this is fixed for future versions, but the 2.2.1 release has the incorrect pinning level.  Luckily there is nothing relevant for Conda in that release, all the changes are related to Bazel-based builds.

Checklist:

  for token resets)


coryan:
ping @conda-forge/google-cloud-cpp

h-vetinari:
Thank you for the speedy handling @coryan! :)

Just to note the issue is not that 2.2.1 _itself_ was broken, but that the run-export of packages built against the original 2.2.0 build only depend on 2.2.*, and therefore pull in 2.2.1 at runtime, leading to missing symbols due to the different namespaces.

As I [mentioned](https://github.com/conda-forge/google-cloud-cpp-feedstock/pull/116#issuecomment-1248397433) on the issue, this could be solved by a repodata patch (retroactively fixing the run export of 2.2.0 down to patch level), or not publishing 2.2.1 at all (but the automerge bot had other ideas). I think the easiest right now is to mark the builds as broken, as you already mentioned that 2.3 is around the corner anyway.

h-vetinari (approved):

xhochy (approved):
Looks like the best way forward for now.

PR #496, opened 2022-10-12
Reset token for r-forge

mbargull:
Copying from cf-staging fails due to "invalid feedstock token".
refs:
- https://github.com/conda-forge/r-forge-feedstock/issues/8

cc @conda-forge/r-forge

----

Checklist:

  for token resets)


mbargull:
Thanks for the merge!
But apparently there's something going on with Travis:
The last runs of https://github.com/conda-forge/admin-requests/actions/workflows/master.yml all show:
```
RuntimeError: Failed to register feedstock token for conda-forge/r-forge-feedstock on travis!
failed to reset token for 'r-forge': CalledProcessError(1, ['conda', 'smithy', 'register-feedstock-token', '--without-circle', '--without-drone', '--feedstock_directory', '/tmp/tmp1m5yxkun/r-forge-feedstock', '--organization', 'conda-forge', '--token_repo', 'https://x-access-token:${GITHUB_TOKEN}@github.com/conda-forge/feedstock-tokens'])
```

beckermr:
We are likely out of requests. Retry the workflow later. This workflow uses the same tokens and code as staged recipes.

dpryan79:
We're still running into invalid feedstock token errors.

beckermr:
Does a retry fix it ?

dpryan79:
I've retried the upload jobs from the master branch already without success. Should I try rerendering as well?

dpryan79:
Rerendering doesn't seem to update anything in this case.

beckermr:
Which feedstocks?

dpryan79:
Just r-forge that I've seen so far: https://github.com/conda-forge/r-forge-feedstock

beckermr:
Ah ok. Looks like Travis is still barfing on the token reset.

beckermr:
I'll have to debug locally but won't get a chance for a while. Is this a big blocker?

dpryan79:
Not a big one, no. It's only blocking a handful of packages in the R 4.2 migration, so it doesn't really matter if this gets addressed today or next week. Thanks!

mbargull:
Apart from the remaining `r-v8` subgraph (about 40 feedstocks), this is now the "biggest" blocker with (if I counted correctly) 6 downstream feedstocks. So, yeah, not many -- but depending on how fast we can handle the `r-v8` stuff, this could get to a "tail end of the R 4.2 migration" status ;).

----
All workflow runs of https://github.com/conda-forge/admin-requests/actions failed since merging this PR with token registration on Travis not working (see https://github.com/conda-forge/admin-requests/pull/496#issuecomment-1275987975 ).
OTOH, registering token for new feedstocks seem to work no problem: https://github.com/conda-forge/staged-recipes/actions/workflows/create_feedstocks.yml

So might be rather specific to https://github.com/conda-forge/r-forge ? (Not sure if relevant/if it helps to pin point the issue: that feedstock also hasn't been built since the CFEP13 token cleanup migration.)

beckermr:
Travis appears to not have this feedstock on its website - https://app.travis-ci.com/github/conda-forge/r-forge-feedstock

IDK why. Maybe we can submit it to staged recipes again to see if that helps?

beckermr:
Or we could disable Travis here...

mbargull:
Thanks for looking into it!
Since the feedstock doesn't use Travis anyway, disabling it would be more than enough :).

beckermr:
Right ok. We'll have to make a special run of the feedstock token reset code or add that feature. I will try for today.

mbargull:
Thank you! Just started it and will check in 4-5 hours (afk).

dpryan79:
Seems to have done the trick!

PR #501, opened 2022-10-25
Mark spdlog builds 1.10.0-*_1 as broken to ABI breakage

traversaro:
The spdlog builds `1.10.0-*_1` have a different ABI w.r.t. `1.10.0-*_0`, resulting in runtime errors if `1.10.0-*_1` is installed with libraries and executables that were builded and linked against `1.10.0-*_0`, see https://github.com/conda-forge/spdlog-feedstock/issues/53 for more info.

Checklist:

  for token resets)

### DIFF
--- a/broken/spdlog.txt
+++ b/broken/spdlog.txt
@@ -1,0 +1,6 @@
+win-64/spdlog-1.10.0-hfbadfc6_1.tar.bz2
+osx-64/spdlog-1.10.0-ha64ae7f_1.tar.bz2
+osx-arm64/spdlog-1.10.0-h6981a3a_1.tar.bz2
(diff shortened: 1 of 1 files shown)


traversaro:
@conda-forge/spdlog if you agree a consensus on https://github.com/conda-forge/spdlog-feedstock/issues/53, we can ping core to get this PR merged.

traversaro:
It seems that the only active members of @conda-forge/spdlog active is @bluescarni that in https://github.com/conda-forge/spdlog-feedstock/issues/53#issuecomment-1291013384 confirmed that this is the way to go. @conda-forge/core the PR is ready for review.

traversaro:
Thanks @beckermr !

PR #509, opened 2022-11-08
Remove pytango 9.4.0rc1 from main channel

beenje:

Checklist:

  for token resets)

I wanted to release a `rc` version of pytango and pushed to a `rc` branch but I did a mistake when adding the `conda_build_config.yaml` file: https://github.com/conda-forge/pytango-feedstock/commit/03f8f7f34de6f59809cd638963aeeb838d89f750

I used `channels_target` instead of `channel_targets` :-( The 9.4.0rc1 version of pytango ended in the main channel
and I'd like to remove it. Or is it possible to move those files to the `pytango_rc` channel?

Fixed in https://github.com/conda-forge/pytango-feedstock/pull/13

ping @conda-forge/pytango


beenje:
Thanks!

PR #514, opened 2022-11-23
Mark libxml2 broken versions

Guspyro:
We identified a bug in a couple of libxml2 versions (2.9.11 and 2.9.12). It caused the _etree.tostring()_ method to return an incorrect xml, with extra closing tags. After creating an [issue](https://github.com/conda-forge/lxml-feedstock/issues/75), they asked me to mark those versions as broken.

Checklist:

  for token resets)


beenje:
@ocefpaf any opinion on this PR?
Should we patch the repodata to avoid installing those libxml2 version with lxml instead?

ocefpaf:
If they are fundamentally broken we need to mark them as so. Let's merge this when the CIs are green.

wolfv:
This is breaking `rviz` installations in the RoboStack project as we've built packages against the `libxml2` versions that are marked as broken. I think that's not the correct approach here.

jakirkham:
There was a fair bit of discussion on this yesterday. Please see PR ( https://github.com/conda-forge/admin-requests/pull/519 ) and links on this

PR #522, opened 2022-12-01
Mark datatoolbox 5.0 as broken (yanked on pypi)

coroa:
Version 5.0 of [datatoolbox](https://gitlab.com/climateanalytics/datatoolbox/) should have been version `0.5.0`
and was only introduced through a misnamed tag and mindless acceptance of the pr update.

It is shadowing /newer/ versions `0.5.[1234]`.

Datatoolbox has a limited number of users within our organisation and is maintained by @geiges and myself. 

We yanked the tag and version from [pypi](https://pypi.org/project/datatoolbox/#history) and would like to remove it from conda-forge as well, thanks.

Checklist:

  for token resets)

ping @conda-forge/datatoolbox

### DIFF
--- a/broken/datatoolbox-v5.0.txt
+++ b/broken/datatoolbox-v5.0.txt
@@ -1,0 +1,1 @@
+noarch/datatoolbox-5.0-pyhd8ed1ab_0.conda


chrisburr:
@beckermr Should we fix this in the bot's metadata as well? https://github.com/regro/cf-graph-countyfair/blob/master/node_attrs/datatoolbox.json#L11

beckermr:
We should yes or the feedstock won't get version updates.

PR #525, opened 2022-12-08
Add broken networkx build

TylerAnderson22:

Checklist:

  for token resets)

Issue (taken from the linked issue below)

> In an empty conda envrionment I run conda install python=3.7 networkx.
I end up getting networkx version 2.7.1.
When I go to the github tag for networkx version 2.7.1 I see that there is a requirement for python >=3.8: https://github.com/networkx/networkx/blob/networkx-2.7.1/setup.py#L195
Looking at commits on networkx-feedstock I see the first commit for 2.7.1 [here](https://github.com/conda-forge/networkx-feedstock/commit/b871075db44054acabcaba96aa046e99d7c39733) and then a correction for the version [here](https://github.com/conda-forge/networkx-feedstock/commit/89d2c0221d2e709c78d0cffc18d50859f1b4b8ff). The correction fixes the dependency issue.
I am getting the first commit, build 0 of networkx 2.7.1. This is unexpected behavior. Is this avoidable on my end or is this something that can be fixed on your end?

Relevant issues: https://github.com/conda-forge/networkx-feedstock/issues/44
ping @conda-forge/networkx-feedstock

### DIFF
--- a/broken/networkx.txt
+++ b/broken/networkx.txt
@@ -1,0 +1,1 @@
+noarch/networkx-2.7.1-pyhd8ed1ab_0.tar.bz2


jakirkham:
Thoughts on this question ( https://github.com/conda-forge/networkx-feedstock/issues/44#issuecomment-1343010770 )?

PR #533, opened 2022-12-26
Marked xeus 3.0.4 as broken

JohanMabille:
the file `xeusConfig.cmake` is missing a search of libuuid for static builds, making it impossible to build a downstream package with this version.


isuruf:
I don't think this needs to be marked broken if there is a new version already

JohanMabille:
Ok, this was just to be sure no one would ping this version or would be hit by the bug before the new version is available on conda-forge.

PR #534, opened 2023-01-04
Mark nbformat 5.1.2 as broken since it was yanked on pypi

### LINKED ISSUE #257: noarch/nbformat-5.1.2-* issue
I wasn't sure how to approach this issue, but from https://github.com/jupyter/nbformat/pull/217, it was shown that the https://anaconda.org/conda-forge/nbformat/files 5.1.2 release was generating inappropriate random ids. We changed 5.1.3 to generate hash ids and yanked the PyPI release. The package isn't broken per se but it's highly problematic. Would the recourse be to mark it as broken for non-technical reasons?

jasongrout:
Fixes #257

We [yanked nbformat 5.1.2 from pypi](https://pypi.org/project/nbformat/#history) shortly after it was released because it made the highly problematic (socially) decision of generating random ids using dictionary words. 5.1.3 corrected this behavior.

## Checklist:

    for token resets)

### DIFF
--- a/broken/nbformat-5.1.2.txt
+++ b/broken/nbformat-5.1.2.txt
@@ -1,0 +1,2 @@
+noarch/nbformat-5.1.2-pyhd8ed1ab_0.tar.bz2
+noarch/nbformat-5.1.2-pyhd8ed1ab_1.tar.bz2


jasongrout:
ping @conda-forge/nbformat - this fixes #257 from a long time ago.

jakirkham:
Thanks Jason and Matt! 🙏

PR #536, opened 2023-01-05
Archive `theano`

jakirkham:
As `theano` is no longer maintained and has been superseded by `aesara`, this archives `theano`.

xref: https://github.com/conda-forge/theano-feedstock/issues/39

## Guidelines for marking packages as broken:

* We prefer to patch the repo data (see [here](https://github.com/conda-forge/conda-forge-repodata-patches-feedstock))
  instead of marking packages as broken. This alternative workflow makes environments more reproducible.
* Packages with requirements/metadata that are too strict but otherwise work are
  not technically broken and should not be marked as such.
* Packages with missing metadata can be marked as broken on a temporary basis
  but should be patched in the repo data and be marked unbroken later.
* In some cases where the number of users of a package is small or it is used by
  the maintainers only, we can allow packages to be marked broken more liberally.
* We (`conda-forge/core`) try to make a decision on these requests within 24 hours.

What will happen when a package is marked broken?

* Our bots will add the `broken` label to the package. The `main` label will remain on the package and this is normal.
* Our bots will rebuild our repodata patches to remove this package from the repodata.
* In a few hours after the `anaconda.org` CDN picks up the new patches, you will no longer be able to install the package from the `main` channel.

## Checklist:

* [ ] I want to mark a package as broken (or not broken):
  * [ ] Make sure your package is in the right spot (`broken/*` for adding the
    `broken` label, `not_broken/*` for removing the `broken` label, or `token_reset/*`
    for token resets)
  * [ ] Added a description of the problem with the package in the PR description.
  * [ ] Pinged the team for the package for their input.

* [x] I want to archive a feedstock:
  * [x] Pinged the team for that feedstock for their input.
  * [x] Make sure you have opened an issue on the feedstock explaining why it was archived.
  * [x] Linked that issue in this PR description.
  * [x] Added links to any other relevant issues/PRs in the PR description.

PR #537, opened 2023-01-05
Archive CI: only use token if needed

jaimergp:
Comes from https://github.com/conda-forge/admin-requests/pull/536#issuecomment-1371999295

Checking the status does not require a token, which is not present or needed on PRs.


jakirkham:
Thanks Jaime! 🙏

Happy to give this a try.

Do we run into any GitHub API limitations when just doing `GET`s without a token? Recall the rate limit was somewhat low when not using a token, but maybe that isn't an issue just for running the occasional test (unless we need to archive a bunch of feedstocks). Any other caveats?

jaimergp:
Thanks @jakirkham - I added a note about possible API rate limiting. I don't think we'll see enough volume here to notice but we'll see. Worst case we do export the public token in a future PR?

I also refactored a bit to ensure that all errors are caught (and not just the first one, which might be annoying for long queues if we ever have to mass archive).

archive_feedstock.py
```diff
@@ -54,14 +56,26 @@ def feedstocks(directory="archive"):
...
-    for feedstock in feedstocks("unarchive"):
-        print(f"Unarchiving {owner}/{feedstock}-feedstock...")
-        archive_repo(owner, f"{feedstock}-feedstock", archive=False, check_only=check_only)
-        time.sleep(0.5)
+    exceptions = []
+    for task in "archive", "unarchive":
+        seen = set()
+        for feedstock in feedstocks(task):
+            verb = "Archiving" if task == "archive" else "Unarchiving"
+            print(f"{verb} {owner}/{feedstock}-feedstock...", end=" ")
```
jakirkham, line 64: It looks like the verb is getting cut short. For [example]( https://github.com/conda-forge/admin-requests/actions/runs/3845806223/jobs/6550384893#step:6:12 ):

```python
Error: ng conda-forge/retworkx-feedstock... [ERROR]
```

jaimergp: I think that's GH's rendering detecting `[ERROR]` and messing with the output?

jaimergp: Let's see if I can use something else.

jakirkham: It might also be that `print`ing is being buffered and so things are getting ordered in an odd way. If we disable buffering, this may help

jaimergp: Added that too just in case!

jakirkham: Interesting, looks like you were right. GHA is interpreting and modifying the line somehow. Using `!!` [works]( https://github.com/conda-forge/admin-requests/actions/runs/3845967004/jobs/6550724600#step:6:10 )

```python
Archiving conda-forge/doesnotexist-feedstock... [!!]
```

jaimergp:
Yay we hit the rate limit 😂 Good news, we don't have to wait and see now.

archive_feedstock.py
jakirkham: Appears [this is intentional]( https://github.blog/2020-09-23-a-better-logs-experience-with-github-actions/#opening-the-door-to-a-more-colorful-experience ). Also looks like there is a way to [affect this behavior]( https://stackoverflow.com/a/71091060 )

jakirkham:
Oh it used to be more like 60 for unauthorized. Maybe that has gone down a lot more? Or we are using more requests than we think?

jaimergp:
The current workflow does work as intended for now, so up to you @jakirkham. We can always refine.

jakirkham:
Thanks Jaime 🙏

Yeah using a GH token seem like the right thing to do. The [default permissions]( https://docs.github.com/en/actions/security-guides/automatic-token-authentication#permissions-for-the-github_token ) are only read for PRs from forks. Though we could tighten it down more if needed.

Am ok with what we have. Though will wait for @beckermr to comment

beckermr:
LGTM!

beckermr:
The standard GHA secrets are safe to use in PRs w/ forks.

PR #539, opened 2023-01-06
[MNT] Remove archive files as they are seen

jakirkham:
Addresses the issue described in comment ( https://github.com/conda-forge/admin-requests/pull/536#issuecomment-1372984066 ). Namely this removes archive files as the repos are (un)archived. This is analogous behavior to what marking (un)broken does.

### DIFF
--- a/archive_feedstock.py
+++ b/archive_feedstock.py
@@ -1,4 +1,5 @@
 import os
+import subprocess
 import sys
 import time
 from pathlib import Path
@@ -43,7 +44,7 @@ def archive_repo(owner, repo, archive=True, check_only=False):
     raise_json_for_status(r)
 
 
-def feedstocks(directory="archive"):
+def feedstocks(directory="archive", check_only=False):
     for path in Path(directory).glob("*.txt"):
         if path.name == "example.txt":
             continue
@@ -52,13 +53,24 @@ def feedstocks(directory="archive"):
                 line = line.strip()
                 if line and not line.startswith("#"):
                     yield line
+        if not check_only:
+            subprocess.check_call(f"git rm {file_name}", shell=True)
+            subprocess.check_call(
+                (
+                    "git commit -m "
+                    f"'Remove {file_name} after {directory[:-1]}ing repos'"
+                ),
+                shell=True
+            )
+            subprocess.check_call("git show", shell=True)
+
 
 
 def main(owner="conda-forge", check_only=False):
     exceptions = []
     for task in "archive", "unarchive":
         seen = set()
-        for feedstock in feedstocks(task):
+        for feedstock in feedstocks(task, check_only=check_only):
             verb = "Archiving" if task == "archive" else "Unarchiving"
             print(f"{verb} {owner}/{feedstock}-feedstock...", end=" ")
             if feedstock in seen:


beckermr (commented):
We need to make sure the feedstocks get archived before we can remove the files I think. What if the api call fails?

jakirkham:
Then an exception is thrown, which would fail the step. So we wouldn't make it to pushing the changes

jakirkham:
Also CI fails as both of the archive files are still around for feedstocks that have been archived

beckermr:
That exception happens after the generator completes. The file deletion is in the generator and so would run no matter what, right?

jakirkham:
The commits would be made yes, but [the (un)archive step]( https://github.com/conda-forge/admin-requests/blob/57b5b69d3d93283e63409696839ba2467ffb38f2/.github/workflows/master.yml#L91-L101 ) should fail (due to the exception). So it won't make it to [the push step]( https://github.com/conda-forge/admin-requests/blob/57b5b69d3d93283e63409696839ba2467ffb38f2/.github/workflows/master.yml#L103-L107 ) as GHA will fail the job. IOW the commits won't go anywhere

Though feel free to correct me if I'm misunderstanding the question

jakirkham:
For example, [this job on `main` is failing in "(un)archive feedstocks"]( https://github.com/conda-forge/admin-requests/actions/runs/3851513604/jobs/6562769078 ). So the "Push changes" step doesn't run. That would be the same case as for this change.

beckermr:
Ah I see. So if some feedstocks worked but others don't, we'd need to manually fix things as we cannot archive twice?

beckermr:
I handled all of these edge cases in the token reset jobs. It retries things that failed by writing the package to a file again. You might just use that?

https://github.com/conda-forge/admin-requests/blob/main/token_reset.py

jakirkham:
Yeah that's true. There is no partial archiving like how there is partial conversion in staged-recipes. Thought about whether to add this, but it seemed like an unlikely problem. Maybe I'm wrong about this though

jakirkham:
Didn't see your last comment when writing that.

Yeah we could. Though maybe we should just soften error handling in cases where a feedstock is already archived

archive_feedstock.py
```diff
@@ -52,13 +53,24 @@ def feedstocks(directory="archive"):
                 line = line.strip()
                 if line and not line.startswith("#"):
                     yield line
+        if not check_only:
```
jaimergp, line 56: Oops, yep I missed this part :D 

With the current exception accumulation we have now, if an error occurs and the repo cannot be archived, we will have deleted the files and committed in the temporarily cloned repo. I guess that's not very problematic in CI usage, but ideally we would only commit on full success.

I don't think it's really important though and if it works most of the time, it should be ok. Just writing it down in case that edge case was not considered.

archive_feedstock.py
```diff
@@ -52,13 +53,24 @@ def feedstocks(directory="archive"):
...
+            subprocess.check_call(f"git rm {file_name}", shell=True)
+            subprocess.check_call(
+                (
+                    "git commit -m "
+                    f"'Remove {file_name} after {directory[:-1]}ing repos'"
```
jaimergp, line 61: Nice trick with `{directory[:-1]}`... I should have used it in `main` too 😂

archive_feedstock.py
jaimergp: Oops, the rest of the conversation hadn't loaded for some reason and I am now realizing you did talk about this. Sorry for the noise 😬

archive_feedstock.py
```diff
@@ -52,13 +53,24 @@ def feedstocks(directory="archive"):
...
+            subprocess.check_call("git show", shell=True)
+
 
 
 def main(owner="conda-forge", check_only=False):
     exceptions = []
     for task in "archive", "unarchive":
         seen = set()
-        for feedstock in feedstocks(task):
+        for feedstock in feedstocks(task, check_only=check_only):
```
jaimergp, line 73: We might need to also get the filename here, not just the line.

jaimergp: Or require a single feedstock per file only?

beckermr:
I removed the files on main causing an issue for now. We can add them back later for testing.

jaimergp:
@jakirkham - if you need me, I have time to work on this! Just let me know if I can / should start pushing to your branch.

jakirkham:
Please feel free Jaime. Thanks for the help 🙏

PR #545, opened 2023-01-09
Reset imageio upload token (c.f. conda-forge/imageio-feedstock#60)

anjos:
Requesting token reset for package imageio following issue conda-forge/imageio-feedstock#1 and advice from @chrisburr at gitter.

### DIFF
--- a/token_reset/imageio.txt
+++ b/token_reset/imageio.txt
@@ -1,0 +1,1 @@
+imageio


anjos:
@chrisburr: this seems to be failing at a different file than the one introduced with this PR.  Is this to be expected?

PR #558, opened 2023-01-09
Reset token for pynldas2

cheginit:

## Checklist:

    for token resets)

### DIFF
--- a/hydrosignatures.txt
+++ b/hydrosignatures.txt
@@ -1,1 +1,0 @@
-hydrosignatures
--- a/token_reset/pynldas2.txt
+++ b/token_reset/pynldas2.txt
@@ -1,0 +1,1 @@
+pynldas2


cheginit:
The token for pynldas2 package needs to be reset.

cheginit:
@beckermr I waited for about an hour and triggered the build, but it's still not fixed. Here is the [link](https://dev.azure.com/conda-forge/feedstock-builds/_build/results?buildId=638525&view=logs&j=47084397-e614-5090-05b8-c04ebacb88b9&t=1f9ba02e-082d-514c-d988-1f2fe7ad8510) to the build.

beckermr:
Just worked. You need a brand new job since azure doesn't pass new secrets to old jobs.

cheginit:
Ah, I see. Makes sense. So, in these situations, I should trigger a new build using, for example, a bot rerender command, or is there a way for me to trigger a new Azure build manually?

beckermr:
PUsh an empty commit to the branch. That's the most reliable.

cheginit:
Great. Thanks!

jakirkham:
Also one can click on the status of the latest commit. On that page there is a "Re-run all checks" button one can push.

Screen Shot 2023-01-09 at 4 25 39 PM

cheginit:
@jakirkham I actually tried that one but didn't work. I think what @beckermr is saying is that in these situations, a simple rerun does not work, and a new build should be triggered.

beckermr:
Yeah many of the buttons we click rerun the same job. An empty commit to main is the safest.

beckermr:
You need a new pipeline instance on azure, not a rerun of an old one. I think those appear with a date and then .1 .2 etc.

cheginit:
I see the same exact thing that you showed. I tried twice with the re-run button, but both times the build failed with the same error.

jakirkham:
It's worth noting that the default setting of the button is to only re-run failed jobs, which won't work (as Matt said above). However there is a dropdown menu on that button that one must use to select re-run all jobs for it to work.

cheginit:
Right, thanks.

PR #572, opened 2023-01-10
reset some more tokens

h-vetinari:
Fix https://github.com/conda-forge/fluidsynth-feedstock/issues/38
Fix https://github.com/conda-forge/gymnasium-feedstock/issues/14 (second attempt)
Fix https://github.com/conda-forge/libmad-feedstock/issues/6

PR #577, opened 2023-01-10
reset some more tokens

h-vetinari:
Fix https://github.com/conda-forge/aws-sdk-cpp-feedstock/issues/659
Fix https://github.com/conda-forge/fluidsynth-feedstock/issues/38 (second attempt)
Fix https://github.com/conda-forge/libmad-feedstock/issues/6 (second attempt)
Fix https://github.com/conda-forge/pygame-feedstock/issues/5


jakirkham:
Is this still needed in light of comment ( https://github.com/conda-forge/admin-requests/pull/572#issuecomment-1376721602 )?

Maybe it is worth restarting the CI on these one more time

h-vetinari:
aws is definitely necessary. The others failed in a different way than the comment you linked (there the validation request succeeded, but the hash was off, which sorted itself out over time). However both https://dev.azure.com/conda-forge/feedstock-builds/_build/results?buildId=638738&view=results and https://dev.azure.com/conda-forge/feedstock-builds/_build/results?buildId=638737&view=results say: 
```
ERROR getting output validation information from the webservice: JSONDecodeError('Expecting value: line 1 column 1 (char 0)')
```

jakirkham:
Maybe let's change this to just `aws` then

h-vetinari:
Or we just rotate tokens once more and see if that fixes it (it's not like we've identified the cause of the issue that made qt-main and gymnasium fail after the first token refresh).

h-vetinari:
So, I've restarted both runs, and they fail exactly as before.

Could we please let the bots do the busy work rather than the humans?

PR #624, opened 2023-01-11
Reset token for hdbpp-es-feedstock

beenje:
The new build of hdbpp-es failed to upload. I guess this is linked to https://github.com/conda-forge/status/issues/137

The feedstock has 2 outputs: `hdbpp-es` and `hdbpp-es-dbg`.

I put both in the same `hdbpp-es.txt` file but I'm not sure this is correct or if I should create two txt files.

### DIFF
--- a/token_reset/hdbpp-es.txt
+++ b/token_reset/hdbpp-es.txt
@@ -1,0 +1,2 @@
+hdbpp-es
+hdbpp-es-dbg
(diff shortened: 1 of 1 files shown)


beenje:
Pipeline failed complaining about `hdbpp-es-dbg-feedstock` so I removed the `hdbpp-es-dbg` sub package from the txt file.

PR #639, opened 2023-01-11
Reset tokens of feedstocks that failed uploads

jakirkham:
Using [this search]( https://github.com/search?q=org%3Aconda-forge+is%3Aissue+is%3Aopen+created%3A%3E%3D2023-01-01++%22invalid+feedstock+token%22+&type=issues&s=created&o=desc ), all of these feedstocks had failed uploads recently. This resets all of their tokens.

## Guidelines for marking packages as broken:

* We prefer to patch the repo data (see [here](https://github.com/conda-forge/conda-forge-repodata-patches-feedstock))
  instead of marking packages as broken. This alternative workflow makes environments more reproducible.
* Packages with requirements/metadata that are too strict but otherwise work are
  not technically broken and should not be marked as such.
* Packages with missing metadata can be marked as broken on a temporary basis
  but should be patched in the repo data and be marked unbroken later.
* In some cases where the number of users of a package is small or it is used by
  the maintainers only, we can allow packages to be marked broken more liberally.
* We (`conda-forge/core`) try to make a decision on these requests within 24 hours.

What will happen when a package is marked broken?

* Our bots will add the `broken` label to the package. The `main` label will remain on the package and this is normal.
* Our bots will rebuild our repodata patches to remove this package from the repodata.
* In a few hours after the `anaconda.org` CDN picks up the new patches, you will no longer be able to install the package from the `main` channel.

## Checklist:

* [ ] I want to mark a package as broken (or not broken):
  * [ ] Make sure your package is in the right spot (`broken/*` for adding the
    `broken` label, `not_broken/*` for removing the `broken` label, or `token_reset/*`
    for token resets)
  * [ ] Added a description of the problem with the package in the PR description.
  * [ ] Pinged the team for the package for their input.

* [ ] I want to archive a feedstock:
  * [ ] Pinged the team for that feedstock for their input.
  * [ ] Make sure you have opened an issue on the feedstock explaining why it was archived.
  * [ ] Linked that issue in this PR description.
  * [ ] Added links to any other relevant issues/PRs in the PR description.

PR #642, opened 2023-01-12
Reset tokens of feedstocks that failed uploads (2)

jakirkham:
After going back through [this search]( https://github.com/search?q=org%3Aconda-forge+is%3Aissue+is%3Aopen+created%3A%3E%3D2023-01-01++%22invalid+feedstock+token%22+&type=issues&s=created&o=desc ) following PR ( https://github.com/conda-forge/admin-requests/pull/639 ), found some more feedstocks that had failed uploads since (or were missed previously). This resets all of their tokens.

<hr>

## Checklist:

    for token resets)

### DIFF
--- a/token_reset/feedstocks_with_issues_2.txt
+++ b/token_reset/feedstocks_with_issues_2.txt
@@ -10,5 +10,6 @@ micropipenv
 mpxj
 nuitka
 openmdao
+openpmd-api
 otter-grader
 pangeo-dask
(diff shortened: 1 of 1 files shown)


jakirkham:
Have restarted CI for all of these

jakirkham:
Should add the search result now looks like this (as opposed to being 10 pages! 😲)

Screen Shot 2023-01-11 at 5 35 45 PM

Also the one feedstock here had a PR merged recently that was buggy. So haven't tried restarting it (though its token has already been reset)

jschueller:
thanks!

jakirkham:
Ofc! :)

PR #657, opened 2023-01-15
Archive trillinos and pytrillinos

hmaarrfk:
They seem rather abbandoned

- https://github.com/conda-forge/trilinos-feedstock/pulls
- https://github.com/conda-forge/pytrilinos-feedstock/pulls

## Checklist:

    for token resets)

### DIFF
--- a/archive/pytrillinos.txt
+++ b/archive/pytrillinos.txt
@@ -1,0 +1,1 @@
+pytrillinos
--- a/archive/pytrillinos.txt
+++ b/archive/pytrillinos.txt
@@ -1,1 +1,1 @@
-pytrillinos
+pytrilinos
--- a/archive/trillinos.txt
+++ b/archive/trillinos.txt
@@ -1,1 +1,1 @@
-trillinos
+trilinos
--- a/archive/trillinos.txt
+++ b/archive/trillinos.txt
@@ -1,0 +1,1 @@
+trillinos


archive/trillinos.txt
```diff
@@ -0,0 +1 @@
+trillinos
```
beckermr, line 1:
```suggestion
trilinos
```

archive/pytrillinos.txt
```diff
@@ -0,0 +1 @@
+pytrillinos
```
beckermr, line 1:
```suggestion
pytrilinos
```

beckermr (commented):
Wrong names.

hmaarrfk:
Well that was embarrassing

beckermr:
Regarding the actual PR, we don't typically archive abandoned feedstocks. Is there a specific reason we should here?

hmaarrfk:
it just really holds up a few tough migrations making it hard to understand how far they actually got.

beckermr:
Hmmmmm ok. I wonder if there is a way we can get the bot to skip them besides archiving.

beckermr:
We could add a feature that lets people have the bot skip their feedstock. I'm not sure how I feel about this.

hmaarrfk:
Hmm ok, lets discuss ideas: https://github.com/conda-forge/conda-forge-pinning-feedstock/blob/main/recipe/migrations/hdf51122.yaml

hmaarrfk:
Opps: ---- > https://github.com/conda-forge/conda-forge.github.io/issues/1724

PR #677, opened 2023-02-15
Remove gazebo macOS builds for which a critical regression was introduced

traversaro:
See https://github.com/conda-forge/gazebo-feedstock/issues/161 for the discussion of the critical regression, that prevents to use gazebo on macOS at all.

## Checklist:

    for token resets)

### DIFF
--- a/broken/gazebo.txt
+++ b/broken/gazebo.txt
@@ -1,0 +1,12 @@
+osx-arm64/gazebo-11.12.0-h3cc2e43_5.conda
+osx-arm64/gazebo-11.12.0-ha73415d_5.conda
+osx-64/gazebo-11.12.0-hd5c0442_5.conda
+osx-64/gazebo-11.12.0-hba8d896_5.conda
+osx-64/gazebo-11.12.0-hab01471_4.conda
+osx-arm64/gazebo-11.12.0-he58534a_4.conda
+osx-arm64/gazebo-11.12.0-h3796533_4.conda
+osx-64/gazebo-11.12.0-h315321b_4.conda
+osx-arm64/gazebo-11.12.0-h24781db_3.tar.bz2
+osx-arm64/gazebo-11.12.0-h74260f6_3.tar.bz2
+osx-64/gazebo-11.12.0-h48b5ba3_3.tar.bz2
+osx-64/gazebo-11.12.0-h0b61fdb_3.tar.bz2


traversaro:
fyi @conda-forge/gazebo

traversaro:
@conda-forge/core the PR is ready for review, thanks!

hmaarrfk:
Does build 6 work? why not just let conda/mamba find build instead of 5? The problem here is that your users won't magically get 6. I'm generally sheepish about these kinds of hotfixes (because there is always a usecase I personally tend to forget) so I like to just "release a new version and fix things there" if that works.

Sometimes it doesn't.

beckermr:
We don’t do hot fixes for bugs in packages generally. We do allow things like this if something was yanked on pypi. That’s a special case since pypi allows this.

traversaro:
I was aware of the bugs policy, but I tought it was a bit different in this case as with this regression (that is a packaging regression, as a patch was accidentally skipped) that is preventing to use the package at all, not just some small bug.

> Does build 6 work? why not just let conda/mamba find build instead of 5? The problem here is that your users won't magically get 6. I'm generally sheepish about these kinds of hotfixes (because there is always a usecase I personally tend to forget) so I like to just "release a new version and fix things there" if that works.

The problem is that between _2 and _6 there were two migrations (ffmpeg and libgdal), see https://github.com/RoboStack/ros-noetic/issues/323#issuecomment-1433873279. Anyhow, we can simply re-create the appropriate builds with the right pin that we need, thanks for the reply!

hmaarrfk:
> The problem is that between _2 and _6 there were two migrations (ffmpeg and libgdal), see https://github.com/RoboStack/ros-noetic/issues/323#issuecomment-1433873279. Anyhow, we can simply re-create the appropriate builds with the right pin that we need, thanks for the reply!

I am sympathetic to this actually. I think this is would fall under a "broken" package to me.

To me, if we find that "the command `mamba install a specific combination of packages`" results in a something totally unusable, that is "broken" to me.

Usually, build number bumps fix this, but when migrations are involved, sometimes the only solution is to patch.

What do yu think @beckermr

beckermr:
The policy is not strict and I'm no expert here. If you all think it should be merged then let's do it.

traversaro:
Thanks @hmaarrfk !

PR #742, opened 2023-06-29
Remove broken kokkos cuda packages

carterbox:
I published CUDA builds for Kokkos 3 targeting arch 3.5 + PTX, but Kokkos does not support CUDA binaries with PTX until Kokkos 4. Conda requires building conda packages for all CUDA compatible major architectures or building for the lowest and including PTX for forward compatability because Conda does not track CUDA archs, however, Kokkos only supports targeting one CUDA arch at build time, so it must be built with PTX.

The effect of leaving these broken packages is that programs that link against them but are running on hardware later than 3.5 may stall/hang OR the runtime will raise an error about the library being compiled for the wrong architecture.

Discussion of this and other Kokkos with CUDA packaging issues are here: 
https://github.com/conda-forge/kokkos-feedstock/issues/21

## Checklist:

    for token resets)

### DIFF
--- a/broken/kokkos3-cuda.txt
+++ b/broken/kokkos3-cuda.txt
@@ -1,0 +1,8 @@
+linux-64/kokkos-3.7.02-h1e7fabd_0.conda
+linux-64/kokkos-3.7.02-h22b3db6_1.conda
+linux-64/kokkos-3.7.02-h66b1f04_0.conda
+linux-64/kokkos-3.7.02-h69fb06d_1.conda
+linux-aarch64/kokkos-3.7.02-h5405a99_0.conda
+linux-aarch64/kokkos-3.7.02-h69b1e34_1.conda
+linux-ppc64le/kokkos-3.7.02-hc023538_1.conda
+linux-ppc64le/kokkos-3.7.02-hd810a49_0.conda


broken/kokkos3-cuda.txt
```diff
@@ -0,0 +1,8 @@
+linux-64/kokkos-3.7.02-h1e7fabd_0.conda
+linux-64/kokkos-3.7.02-h22b3db6_1.conda
+linux-64/kokkos-3.7.02-h66b1f04_0.conda
+linux-64/kokkos-3.7.02-h69fb06d_1.conda
+linux-aarch64/kokkos-3.7.02-h5405a99_0.conda
+linux-aarch64/kokkos-3.7.02-h69b1e34_1.conda
+linux-ppc64le/kokkos-3.7.02-hc023538_1.conda
+linux-ppc64le/kokkos-3.7.02-hd810a49_0.conda
```
jakirkham, line 8: Do both `build/numbers` need to be marked broken or only the `0` ones?

```suggestion
linux-64/kokkos-3.7.02-h1e7fabd_0.conda
linux-64/kokkos-3.7.02-h66b1f04_0.conda
linux-aarch64/kokkos-3.7.02-h5405a99_0.conda
linux-ppc64le/kokkos-3.7.02-hd810a49_0.conda
```

jakirkham:
Thanks Daniel! 🙏

Had a question above

broken/kokkos3-cuda.txt
carterbox: Yes. All build numbers for this version are affected.

jakirkham: Oh ok. Thought for some reason PR ( https://github.com/conda-forge/kokkos-feedstock/pull/24 ) was fixing this issue. Sounds like I misunderstood

PR #755, opened 2023-07-11
Mark arch versions of `allure-*` packages as broken

kenodegard:
The initial implementation of the allure-python-feedstock (for the `allure-*` packages) mistakenly built for all archs (instead of being noarch), see https://github.com/conda-forge/allure-python-feedstock/pull/2.

This "removes" these redundant packages.

## Checklist:

    for token resets)

Xref https://github.com/conda-forge/allure-python-feedstock/pull/2
Cc @conda-forge/allure-python


jakirkham:
So these packages are not actually broken right? Just extra?

If so, think we can leave them there

nsoranzo:
I'm just worried conda may prefer older platform-specific packages to newer noarch ones, but if that's not the case I suppose this is not needed.

ocefpaf:
> I'm just worried conda may prefer older platform-specific packages to newer noarch ones.

I never ran solver experiments to check this but, in theory, the solvers based on libmamba always prioritize higher build numbers. If the noarch are newer, they'll be preferred.

kenodegard:
Ok I'll close this for now, can always reopen if we find that these arch packages cause issues later

Issue #752, opened 2023-07-05
Adding access control to CI providers and other resources


jaimergp:
This comes from https://github.com/conda-forge/conda-forge.github.io/issues/1875

---

I've been thinking about how to implement opt-in CI access control for Travis and other services we might have in the future. I think a potential solution might be formulated like this:

* We will open a new `resource-access/` directory (or something like that)in this repository.
* This directory will contain a number of files (e.g. TXT files), one per resource where we are providing opt-in mechanisms:
* Each TXT file will simply list which feedstocks should have access to that resource, similar to what we do with the osx-arm64 migration file.

For example:

```
resource-access/
  travis_ci.txt
  a_gpu_provider.txt
  some_aws_credits.txt
  long_running_jobs.txt
```

```
# travis_ci.txt
numpy
scipy
```

```
# long_running_jobs.txt
tensorflow
pytorch
```

The workflow would be something like this:

1. A feedstock maintainer opens a PR adding their feedstock to the desired service file. Different services might need different application requirements; these could be added as comments at the top of the file or something.
2. The core team will review the application and merge if accepted.
3. Once merged, automation machinery will get the diff of the new change and register the feedstock for the new service. Each service might require different automation.
4. A cronjob will run every now and then to make sure that all listed feedstocks have access to the service, and will fix it if needed.

The cronjob might be tricky to set up so an alternative it's two have a three file set-up for each service, similar to what we do with the broken/not_broken infra:

* `the_service.txt` -- to be modified by the automation infrastructure, not the user; maybe in a different repository to make sure human errors are prevented. This is here for our own record keeping so we know who has access to what service, quickly.
* `grant-access/the_service.txt` -- the input file to signal the automation to add this feedstock to the service
* `revoke-access/the_service.txt` -- the input file to signal the automation to remove access to this feedstock.

Thoughts?

PR #767, opened 2023-07-26
Opt-in resources request mechanism

jaimergp:
WIP. See https://github.com/conda-forge/admin-requests/pull/767#issuecomment-1695343785

Closes https://github.com/conda-forge/admin-requests/issues/752.


access_control.py
```diff
@@ -0,0 +1,221 @@
...
+
+def process_access_control_requests():
+    """Process access control requests"""
+    print("Processing access control request")
+    _process_access_control_requests("grant_access", remove=False)
+    _process_access_control_requests("revoke_access")
+
+
+def _process_request_for_feedstock(feedstock, resource, remove, policy_args):
+    feedstock_clone_path = f"/tmp/{feedstock}"
```
jaimergp, line 75: We should use stdlib's `tempfile` here instead.

grant_access/example.txt
```diff
@@ -0,0 +1,3 @@
+# filename must match the target resource
+# file contents: feedstock name(s), sans -feedstock, one per line
+foo-feedstock
```
jaimergp, line 3:
```suggestion
foo
```

For consistency with the other parts of `admin-requests`, this should work _without_ the -feedstock suffix. Can you double check this is true? Thanks

aktech:
This PR adds the ability to request grant/revoke access to an Opt-in ci resource (cirun in this case).

This depends on https://github.com/conda-forge/conda-smithy/pull/1703/, which adds this functionality in the `conda-smithy` CLI, See this https://github.com/conda-forge/conda-smithy/pull/1703#issuecomment-1695300650

### Requesting or Revoking CI Resources Access

For specific CI resources that are available on an opt-in basis, you can request access by submitting a PR with your feedstock name added to a relevant file under the `grant_access/` directory. If you wish to revoke access, please make changes in the `revoke_access/` directory.

List of opt-in resources:

- **GPU CI**:
  - `cirun-gpu-runner.txt`: Grants access to the GPU runner on the main branch only.
  - `cirun-gpu-runner-pr.txt`: Provides GPU runner access for pull requests as well.
 
more of these will be added later, this is to get us started.

We're using txt files, for this via PR mechanism for the following reasons:
- consistency with other mechanisms like broken/not-broken, archive/un-archive
- dumb method for people to request/revoke access, with low possibility of errors/issues. We're eventually updating the `.access.yml` (to be created) file in the `.cirun` repo ([example](https://github.com/cirunlabs/.cirun/edit/main/.access.yml)) eventually, this txt file method with concurrency one, also avoids creating conflicts in the `.access.yml` file.

environment.yml
```diff
@@ -8,3 +8,9 @@ dependencies:
   - requests
   - conda-smithy
   - pygithub
+#  - cirun
+#  - conda-smithy
+# TEMP Remove pip install after testing
+  - pip:
+    - git+https://github.com/jaimergp/conda-smithy.git@gha-self-hosted-cirun
+    - git+https://github.com/aktechlabs/cirun-py.git@main
```
jaimergp, line 16: We need to release these two before merging. Can you get started with `cirun-py` or are you expecting more changes there?

access_control.py
```diff
@@ -0,0 +1,225 @@
...
+import sys
+import tempfile
+from pathlib import Path
+
+import requests
+from ruamel.yaml import YAML
+
+GH_ORG = os.environ.get("GH_ORG", "conda-forge")
+
+CIRUN_FILENAME_RESOURCE_MAPPING = {
```
jaimergp, line 17: This is the configuration dictionary that maps `{grant,revoke}_access/<name>.txt` to an actual resource in ... `.access_control.yml`? Or where? Can you document the expected schema with a `#: comment` on top?

access_control.py
```diff
@@ -0,0 +1,225 @@
...
+        "resource": "cirun-gpu-runner",
+        "policy_args": ["pull_request"]
+    }
+}
+
+
+ACCESS_YAML_FILENAME = ".access_control.yml"
+
+
+def _get_access_control_files(path):
```
jaimergp, line 31: Let's add simple docstrings and minimal type hints to all functions. This is the "get the input files" function, right?

access_control.py
```diff
@@ -0,0 +1,225 @@
...
+
+
+def _get_filename_feedstock_mapping(path):
+    access_files = _get_access_control_files(path)
+    file_name_feedstock_mapping = {}
+    for file_path in access_files:
+        with open(file_path, "r") as f:
+            feedstocks = f.readlines()
+            feedstocks = [feedstock.strip() for feedstock in feedstocks]
+            file_name = file_path.parts[-1].strip(".txt")
```
jaimergp, line 46: This should be its own function I think (parse an individual input file)

access_control.py
```diff
@@ -0,0 +1,225 @@
...
+            file_name_feedstock_mapping[file_name] = feedstocks
+    return file_name_feedstock_mapping
+
+
+def _process_access_control_requests(path, remove=True):
+    print(f"Processing access control requests for {path}")
+    grant_access_request = _get_filename_feedstock_mapping(path)
+    for filename, feedstocks in grant_access_request.items():
+        if filename in CIRUN_FILENAME_RESOURCE_MAPPING:
+            resource_mapping = CIRUN_FILENAME_RESOURCE_MAPPING.get(filename)
```
jaimergp, line 56: Simplify to:

```suggestion
        resource_mapping = CIRUN_FILENAME_RESOURCE_MAPPING.get(filename)
        if resource_mapping:
```

access_control.py
```diff
@@ -0,0 +1,225 @@
...
+            resource = resource_mapping.get("resource")
```
jaimergp, line 57: Can this be absent? Or is it always expected?

access_control.py
```diff
@@ -0,0 +1,225 @@
...
+    for file_path in access_files:
+        with open(file_path, "r") as f:
+            feedstocks = f.readlines()
+            feedstocks = [feedstock.strip() for feedstock in feedstocks]
+            file_name = file_path.parts[-1].strip(".txt")
+            file_name_feedstock_mapping[file_name] = feedstocks
+    return file_name_feedstock_mapping
+
+
+def _process_access_control_requests(path, remove=True):
```
jaimergp, line 51: Why `remove=True`? I would expect requests to _grant_ access be more common than those _revoking_. I'd change the default to False.

access_control.py
```diff
@@ -0,0 +1,225 @@
...
+                )
+                action = "remove" if remove else "add"
+                update_access_yaml(resource, feedstock_repo, action=action)
+
+
+def process_access_control_requests():
+    """Process access control requests"""
+    print("Processing access control request")
+    _process_access_control_requests("grant_access", remove=False)
+    _process_access_control_requests("revoke_access")
```
jaimergp, line 73: Be explicit about the remove kwarg here too.

access_control.py
```diff
@@ -0,0 +1,225 @@
...
+
+
+def _process_request_for_feedstock(feedstock, resource, remove, policy_args):
+    feedstock_clone_path = tempfile.mkdtemp()
```
jaimergp, line 77: Do not forget to delete when done.

access_control.py
```diff
@@ -0,0 +1,225 @@
...
+    clone_cmd = f"git clone --depth 1 https://github.com/{GH_ORG}/{feedstock}.git {feedstock_clone_path}"
+    print(f"Cloning: {clone_cmd}")
+    subprocess.run(
+        clone_cmd,
+        shell=True,
+    )
+
+    register_ci_cmd = [
+        "conda-smithy register-ci",
+        f"--organization {GH_ORG}",
```
jaimergp, line 88:
```suggestion
        f"--organization={GH_ORG}",
```
Or separate it in two tokens. But this will fail if separated by a space because it will be taken as a single command.

jaimergp: Ah no wait, you are joining everything after. Hm, in that case why create a list to begin with?

access_control.py
```diff
@@ -0,0 +1,225 @@
...
+    subprocess.check_call(register_ci_cmd_str,
+        shell=True,
+    )
+
+
+def check_if_repo_exists(feedstock_name):
+    repo = f"{feedstock_name}-feedstock"
+    owner_repo = f"{GH_ORG}/{repo}"
+    print(f"Checking if {owner_repo} exists")
+    response = requests.get(f"https://api.github.com/repos/{owner_repo}")
```
jaimergp, line 117: To save API calls (I know 😂) we usually just query the regular website, as in:

```suggestion
    response = requests.get(f"https://github.com/{GH_ORG}")
```

access_control.py
```diff
@@ -0,0 +1,225 @@
...
+    _check_for_path("revoke_access")
+
+def _commit_after_files_removal(push=True):
+    subprocess.check_call(
+        "git add .",
+        shell=True,
+    )
+
+    subprocess.check_call(
+            "git commit --allow-empty -am 'Remove access control files'",
```
jaimergp, line 143: What's the reason behind --allow-empty? Are we expecting no files to be added? Can we check for that and save the commit just in case we don't have to do anything?

access_control.py
```diff
@@ -0,0 +1,225 @@
...
+    print(f"Updating {filename}")
+    yaml = YAML()
+    yaml.indent(mapping=2, sequence=4, offset=2)  # Indent settings for 2-space mapping
+
+    with open(filename, 'r') as f:
+        content = yaml.load(f)
+
+    # If resource doesn't exist, create it
+    if resource not in content['access_control']:
+        content['access_control'][resource] = []
```
jaimergp, line 186:
```suggestion
    content['access_control'].setdefault(resource, [])
```

access_control.py
```diff
@@ -0,0 +1,225 @@
...
+
+    if action == "add":
+        # Check if feedstock already exists
+        if any(entry["feedstock"] == feedstock_name for entry in content['access_control'][resource]):
+            print(f"Feedstock {feedstock_name} already exists under {resource}. Skipping addition.")
```
jaimergp, line 191: Is there a scenario where a feedstock can be added to the same resource twice, maybe with different policies? 🤔 I cannot think of anything but double checking.

.access_control.yml
jaimergp, line 1: We need to document this schema somewhere and validate it in a minimal way.

access_control.py
```diff
@@ -0,0 +1,225 @@
...
+            return
+
+        entry = {"feedstock": feedstock_name}
```
jaimergp, line 194: Your example `.access_control.yml` shows entries with a policy key too, but this line hardcodes that item to a single-key dictionary. How are we handling policies then?

access_control.py
```diff
@@ -0,0 +1,225 @@
...
+        content['access_control'][resource].append(entry)
+
+    elif action == "remove":
+        if not content['access_control'][resource]:
+            print(f"No feedstock found under resource {resource}.")
+            return
```
jaimergp, line 200: This should be an error I think?

access_control.py
aktech: I forgot to remove the policies key from the `.access_control`, we do not have the policy name here in `admin-requests`. My assumption is it won't be tracked in `.access_control`.

access_control.py
aktech: I think accidentally left the `allow-empty`

access_control.py
aktech: Only for readability purpose

access_control.py
aktech: This is always expected, I have removed the `.get` and now directly accessing it.

isuruf (commented):
AFAICT, a user adds a `grant_access/cirun-gpu-runner.txt` file with the feedstock name in it and then the action here reads it, grants access and adds it to `.access_control.yml` and deletes `grant_access/cirun-gpu-runner.txt`.

This is a bit strange and different from other mechanisms where the name of the file does not matter and there are no conflicts because people are using different file names.

jaimergp:
Oh, that's an excellent point. That's going to cause churn. Perhaps we need to change the system so we have a `resource<space>feedstock` line syntax, or at this point just require a simple YAML?

isuruf:
Or just subfolders. `grant_access/cirun-gpu-runner/example.txt`

revoke_access/gpu-runner-pr/example.txt
```diff
@@ -0,0 +1,3 @@
+# filename must match the target resource
```
isuruf, line 1:
```suggestion
```

isuruf:
@jaimergp, @aktech, this is good to go right?

(We need https://github.com/conda-forge/conda-smithy/pull/1793 to clean this PR a little bit and https://github.com/conda-forge/conda-smithy/pull/1794 to handle the teams correctly, but it should be possible to test without them)

README.md
```diff
@@ -53,3 +53,16 @@ If you want to request a feedstock to be unarchived, follow the same steps but f
...
+Certain CI resources are opt-in only. If you want to request access to these resources, please
+submit a PR adding your feedstock name to an appropriately-named file in `grant_access/`. Conversely, to revoke
+access to a certain resource, use the `revoke_access/` directory.
+
+Available opt-in resources:
+
+- Travis CI: `travis.txt`
+- GPU CI:
+  - `cirun-gpu-runner.txt`: access to GPU runner on main branch only.
+  - `cirun-gpu-runner-pr.txt`: access to GPU runner on pull requests as well.
```
jaimergp, line 68: These docs are out of date I think. We are using directory names now.

jaimergp:
> this is good to go right?

I think so. There might be typos, mismatching identifiers, or simple errors that we will easily reveal once we try this live. The docs (as per comment above) seem to be outdated too, but I suggest we amend them once we know it's working. 

One possible pain point is the maintenance of the access_control YAML here (for bookkeeping), because it might become a conflict source across PRs, so we need to ensure it's written in a consistent way (sort keys, etc). Again the kind of thing we will easily observe live.

isuruf:
I ran this on a different branch and fixed a bunch of failures.
Need https://github.com/conda-forge/conda-smithy/pull/1799 and https://github.com/conda-forge/conda-smithy/pull/1796

conda-forge-daemon:
This is ready for a review now.

isuruf:
This is ready for a review now.

access_control.py
```diff
@@ -0,0 +1,428 @@
...
+            subprocess.check_call(
+                [
+                    'conda', 'smithy', 'generate-feedstock-token',
+                    '--unique-token-per-provider',
+                    '--feedstock_directory', feedstock_dir,
+                    *owner_info,
+                ]
+            )
+
+            print("Register new feedstock token with provider and feedstock-tokens repo.")
```
beckermr, line 251: We need to double check that smithy, our tokens repo, and the webservices are properly interacting with the unique tokens. I don't recall how much testing I have done.

isuruf: I checked and it works as intended. (At least for github actions as the provider)

beckermr: Ah just so I am clear. You looked in the tokens repo, the json was formatted correctly, and the tokens on the other CI providers we did not touch still work?

isuruf: Yes to both.

beckermr: great! thank you for making sure all of the details worked!

beckermr (approved):
LGTM!

isuruf:
Thanks for the review. I'll start some docs PRs.

jaimergp:
Awesome! Thanks! Please ping me in the docs PR when needed!

PR #774, opened 2023-07-28
Remove Quarto packages affected by packaging bug

mfisher87:
Hello, I'm sad to request a large number of Quarto packages to be removed from `conda-forge`. 

We found that a patch was causing unintended (and very bad) behavior when using Quarto to install TinyTeX. The intent of the patch was to enable the Quarto conda package to install TinyTex into the current conda environment instead of to the path Quarto normally uses, `~/.TinyTex`.  The unintended result of the patch was that `quarto install tinytex` **overwrites** the entire conda environment with a TinyTeX installation.

A discussion between feedstock maintainer and Quarto devs yielded a decision to remove that patch. Quarto's own packages and conda-forge packages will all look for TinyTeX in, and install TinyTeX to, the same `~/.TinyTeX` location. Versions `1.3.433-1` and `1.3.427-2` were released to fix the latest versions.

Long-term, we [want to enable the Quarto conda package to get its TinyTeX dependency from `conda-forge`](https://github.com/conda-forge/quarto-feedstock/issues/26).

For discussion on marking these packages broken by packaging maintainers and Quarto devs, start here:

https://github.com/conda-forge/quarto-feedstock/issues/25

:heart:  Thank you for your time!

## Checklist:

    for token resets)


mfisher87:
@conda-forge/core is there anything I can do to help make a decision on this request?

Thanks for your time :)

PR #787, opened 2023-08-10
mark libcxx 17.0.0.rc1 as broken on osx

h-vetinari:
Needs fixing because it's messing up clang (only in the rc-label, but still), as it still pulls in 17.0.0.rc1 on a MacOS 10.9 runtime even though it hard a run-constraint >=10.13.

### DIFF
--- a/broken/libcxx.txt
+++ b/broken/libcxx.txt
@@ -1,0 +1,1 @@
+osx-64/libcxx-17.0.0.rc1-hb5c98a3_0.conda


h-vetinari:
@beckermr, CI here fails because only checks the main channel. Should this PR still work to mark a build under a label (here `llvm_rc`) as broken? If not, how could I do that?

beckermr:
You can mark it by hand on anaconda.org. I can add you to our account since you are in core.

h-vetinari:
> You can mark it by hand on anaconda.org. I can add you to our account since you are in core.

Cool thanks! I don't think I have an account on anaconda.org (and it doesn't look like it supports signup via github) - I guess I need to create one?

beckermr:
Yep. Create one and dm your username.

h-vetinari:
Thanks! I'll close this PR then.

PR #792, opened 2023-08-13
archive neworder

virgesmith:
see https://github.com/conda-forge/neworder-feedstock/issues/28

## Checklist:

    for token resets)

### DIFF
--- a/archive/repo.txt
+++ b/archive/repo.txt
@@ -1,0 +1,1 @@
+neworder


h-vetinari:
Any rationale for archiving? (I don't see one in the other issue either)

I don't know the history of this feedstock, and I'm not against archiving of course, just curious what's up.

jaimergp:
Looks like it's more of a project decision than anything else: https://github.com/virgesmith/neworder/issues/81 & https://github.com/conda-forge/neworder-feedstock/pull/27#issuecomment-1567108361

I don't think we should archive it just because the upstream maintainers don't want to invest energy in it (which is totally valid, don't get me wrong). @virgesmith you are free to remove yourself from the maintainers in the feedstock; maybe someone will take over. I don't know if we have a precedent for an "empty" maintainers list but worst case we can add a filler.

ocefpaf:
> I don't know if we have a precedent for an "empty" maintainers list but worst case we can add a filler.

We don't. Usually a core members steps in for abandoned packages. I'll add myself just in case and the current maintainers can remove themselves.

virgesmith:
Thanks @jaimergp and @ocefpaf, I've removed myself. I don't have the bandwidth to maintain conda packages (as you suspected) but it's still on pypi. I don't think there's any great need for it to continue be maintained here, but happy for you to take over if you so wish

PR #794, opened 2023-08-15
Mark all versions of tabmat v4.0.0dev1 as broken

MartinStancsicsQC:

## Checklist:

    for token resets)

## Description

4.0.0.dev1 was supposed to be a pre-release for the 4.0 version of the tabmat package, but due to a mistake in updating the recipe, it was not actually marked as a pre-release. As a result, any downstream dependency not pinning <4.0 are pulling this pre-release into their environments, leading to broken environments/CI pipelines.


MartinStancsicsQC:
Yes, it should have been published under the `tabmat_dev` label. If there is no mechanism for changing the label then marking the release as broken (very recent, removing it should not break anyones environments) should be okay, too. We can always re-release it under the proper label in the future.

xhochy:
> https://github.com/orgs/conda-forge/teams/core, have we ever changed labels "on the spot" on published packages like this?

We have done this years ago but since we have automation nearly everywhere, I haven't seen us doing this manually.

@MartinStancsicsQC I think you're fine with just uploading a new build? Then I would merge here.

PR #798, opened 2023-08-16
Req to remove rc build from main iris channel.

pp-mo:
The problem was created by a malformed push to the iris-feedstock/release-candidate branch.
This has mistakenly created an RC version package in the iris "main"  channel.

## Checklist:

    for token resets)

### DIFF
--- a/broken/iris_main_remove_v3x0x0rc0.txt
+++ b/broken/iris_main_remove_v3x0x0rc0.txt
@@ -1,0 +1,1 @@
+noarch/iris-3.7.0rc0-pyha770c72_0.conda


pp-mo:
Thanks for this @ocefpaf 
I hope this avoids too many people downloading it "by mistake", but I can already see quite a few downloads.
I guess there is no way of controlling this.  Presumably the artifact is not _removed_ from the channel, that is policy ??

jakirkham:
Users should see the stable version get picked up when updating the package specifically or updating all packages in an environment

We prefer moving things to a `broken` label instead of deletion. This is helpful for a number of reasons

1. It provides a record
2. Allows for reproducibility

1 in particular is useful for the users who picked up this package as they can see we marked it broken (so it is accidental that they picked up the package before)

PR #802, opened 2023-08-24
Remove broken `bw2calc` (label `dev`) Packages

michaelweinold:

## Checklist:

    for token resets)

## Description

The `dev` versions of [`bw2calc`](https://github.com/brightway-lca/brightway2-calc) were added to the conda-forge channel under the `dev` label. Unfortunately, the package was not released as `noarch`. This complicates downstream development efforts depending on this package. Related discussions in these PRs:

- https://github.com/conda-forge/bw2calc-feedstock/pull/14
- https://github.com/emscripten-forge/recipes/pull/616

The platform specific `dev` versions (cf. the `broken/bw2calc_broken.txt` file) should be removed entirely. They are to be replaced with a `noarch` version.

🔔 @conda-forge/bw2calc


michaelweinold:
@conda-forge/core, this is ready for review. Not sure why the checks are failing. I formatted the file exactly like eg. https://github.com/conda-forge/admin-requests/pull/686/files, where the checks are passing.

ocefpaf:
I wonder if it is the fact that those are in the dev label that is causing this error. I'm not sure our automated labeling system covers that case.

michaelweinold:
> Also, the channel should not be used in general but artefacts should use their own <name>_dev channel for development releases. Otherwise, an enduser cannot select from which artefact they want to try out dev releases.

Are you saying we should have created a branch `bw2calc_dev` instead of `dev` in the `bw2calc` repo?

We successfully installed the package by specifying the channel+label in the package specs of the usual environment files:

```
- https://conda.anaconda.org/conda-forge/label/bw2calc_dev::bw2calc=2.0.dev14
```

Maybe I'm not following - in any case, we tried to follow [CFEP-05](https://github.com/conda-forge/cfep/blob/main/cfep-05.md) to the letter.

h-vetinari:
Yes, the automation here only works for `main`. For marking things in different labels as broken it needs someone from core (and in the cf-org on https://anaconda.org/conda-forge) to mark them broken by hand.

h-vetinari:
The label is [already](https://anaconda.org/conda-forge/bw2calc/files) `bw2calc_dev`, as it should be.

> Unfortunately, the package was not released as `noarch`. This complicates downstream development efforts depending on this package.

Now that you have a noarch build (`2.0.dev14`), why do you still need to mark `2.0.dev13` as broken? The solver should already pull in the newer one.

PR #805, opened 2023-08-30
Remove broken kedro-pack-1.0.0

millsks:

## Checklist:

    for token resets)

### DIFF
--- a/broken/kedro-pack.txt
+++ b/broken/kedro-pack.txt
@@ -1,0 +1,1 @@
+noarch/kedro-pack-1.0.0-pyhd8ed1ab_0.conda


millsks:
ping @conda-forge/kedro-pack please review.

beckermr:
What is broken here? I don't see a description.

millsks:
Hi Matthew. This version is missing the actual source that is goes along with it.  The version ingredient also changed so to get the correct version someone would have to specify kedro-pack==0.18.12.1 because the newer version is lower than the 1.0.x.  It would choose the wrong version every time.

beckermr:
thank you!

millsks:
Thanks @beckermr!  What is the normal turn around where we would see this version removed from conda-forge?

beckermr:
in the next 20-40 minutes or so - the repodata patches rebuild and then a cdn sync happens

PR #812, opened 2023-09-08
Mark soupsieve 2.5_0 as broken

igortg:
Mark soupsieve 2.5 build 0 as broken since it doesn't work on Python 3.6 environments (Build 1 fix this by adding `python >=3.8` to the recipe).

## Checklist:

    for token resets)


igortg:
Ping @conda-forge/soupsieve

PR #830, opened 2023-09-27
Mark openturns 1.21rc1 as broken

jschueller:
openturns 1.21rc1 is pulling a wrong version of dlib

Will close https://github.com/conda-forge/openturns-feedstock/issues/178


ocefpaf:
Usually we recommend a repo data patch in these cases but b/c this is an rc release I'll merge this. Note that, if you change your mind, you can mark it as "unbroken" and create the patch.

jschueller:
thanks!

PR #853, opened 2023-10-27
Create `token_reset/sccache.txt`

ajschmidt8:

## Checklist:

    for token resets)

### DIFF
--- a/token_reset/sccache.txt
+++ b/token_reset/sccache.txt
@@ -1,0 +1,1 @@
+sccache


ajschmidt8:
This PR is a follow-up to:

- https://github.com/conda-forge/admin-requests/pull/837
- https://github.com/conda-forge/admin-requests/pull/838

ajschmidt8:
@ocefpaf, thanks for merging. do you know why the token reset fails for Travis?

see these latest automated commits to `main`:

- https://github.com/conda-forge/admin-requests/commit/1ecdd996c5fa6bcb51ccd14dfc296f8ffe3fc365
- https://github.com/conda-forge/admin-requests/commit/c1e7950abe2d71cfa71ac110386e4ea2379d1c78
- https://github.com/conda-forge/admin-requests/commit/072786d6f80ff1bb7c1b80b87202ee7976a03900

ocefpaf:
Nope but you can try the no Travis option. Not sure if that is documented (I'm on mobile only to check) but you can find examples in the past PRs merged here.

jakirkham:
Yeah skipping Travis would be easier

Unfortunately it doesn't sound like an option in this case

Relevant past discussion: https://github.com/conda-forge/admin-requests/pull/838#issuecomment-1748021112

ajschmidt8:
We need the Travis token to build arm variants of the sccache feedstock.

Is there no one who can troubleshoot this problem?

I would assume this problem will affect all Travis tokens across conda-forge which seems like a pretty big problem.

The alternative of configuring every feedstock to do cross-compilation for arm variants seems like a lot more effort than just fixing the Travis token issue.

PR #860, opened 2023-11-13
Bump to Python 3.9

jaimergp:
`conda-forge-repodata-patches` uses `typing.Annotated`, which needs Python 3.9. This has been failing for the last 3 months 😬

PR #861, opened 2023-11-13
Bump to Python 3.10

jaimergp:
Follow up for #860. Type `|` unions require Python 3.10.


beckermr:
If only they stopped changing things and adding fancy stuff - It'd just work!

PR #874, opened 2023-11-29
Allow Magma feedstocks to use GPU runners

carterbox:

## Checklist:

    for token resets)

* [x] I want to build a package on GPU runners

### DIFF
--- a/grant_access/gpu-runner-pr/magma.txt
+++ b/grant_access/gpu-runner-pr/magma.txt
@@ -1,0 +1,1 @@
+libmagma
--- a/grant_access/gpu-runner/magma.txt
+++ b/grant_access/gpu-runner/magma.txt
@@ -1,1 +1,0 @@
-libmagma
--- a/grant_access/gpu-runner/magma.txt
+++ b/grant_access/gpu-runner/magma.txt
@@ -1,0 +1,1 @@
+libmagma


jakirkham:
cc @jaimergp @isuruf

carterbox:
https://github.com/Quansight/open-gpu-server/pull/16

grant_access/gpu-runner-pr/magma.txt
```diff
@@ -0,0 +1 @@
+libmagma
```
isuruf, line 1: Just checking. Do you want to build PRs too?

carterbox: Yes, please. Why not?

grant_access/gpu-runner/magma.txt
```diff
@@ -0,0 +1 @@
+libmagma
```
isuruf, line 1: You don't need both these files

carterbox:
It was unclear to me from the README that gpu-runner-pr implies gpu-runner. i.e. That the permissions granted to a feedstock in gpu-runner are a subset of those granted to a feedstock in gpu-runner-pr.

jaimergp:
Thanks @carterbox! I'll make sure to document this nicely.

jakirkham:
Maybe this is too wordy. Perhaps this could be

* `gpu-runner-branches-only`
* `gpu-runner-branches-and-prs`

Or some other more succinct option

(assuming I even understood the intent correctly 😅)

jaimergp:
Added https://github.com/conda-forge/admin-requests/pull/886

PR #892, opened 2023-12-06
Mark libpq and postgresql 16.1 builds 1 2 and 4 as broken due to openssl 3.2

hmaarrfk:
See discussion https://github.com/conda-forge/postgresql-feedstock/issues/191

cc: @h-vetinari @akrherz 

These packages were built requiring "openssl>=3.2", but with that version of openssl, users experience errors at runtime because we forgot to merge in a patch

the fix is proposed: https://github.com/conda-forge/postgresql-feedstock/pull/192/files

## Checklist:

### DIFF
--- a/requests/libpq-openssl32.yml
+++ b/requests/libpq-openssl32.yml
@@ -18,3 +18,114 @@ packages:
   - win-64/libpq-16.1-h43585b0_1.conda
   - win-64/libpq-16.1-h43585b0_2.conda
   - win-64/libpq-16.1-h1beaf6b_4.conda
+
+  # postgresql builds 1, 2, and 4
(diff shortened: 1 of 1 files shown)


hmaarrfk:
cc: @conda-forge/postgresql

akrherz (approved):
Thank you

h-vetinari:
I think we need to mark all dependent builds from the same run as broken too. E.g. `postgresql` depends exactly on `libpq` and so those builds would run into weird resolver errors. Same for `postgresql-python`

hmaarrfk:
Ok thanks for the suggestion this is indeed a good idea.

PR #896, opened 2023-12-15
ENH remove conda-build 3.28.1 due to performance regression

beckermr:

## Checklist:

conda-build 3.28.1 has a performance regression related to https://github.com/conda/conda-build/issues/5106 and its fix. Anaconda inc is seeing the same thing. 

cc @conda-forge/conda-build


mbargull:
Fine by me.
To give further context: This regression also leads to some broken builds on Travis CI (build log limit exceeded) and can even cause builds to timeout on Azure Pipelines (esp. Windows where +5h build times have been seen).

mbargull (approved):

mbargull:
A new `conda-build=3.28.2` with this fixed is expected to land beginning of next week, hence we could try out new `3.28` builds soon anyway, so nothing much lost by going with `3.27` until then.

PR #899, opened 2023-12-21
Mark micromamba 2 alpha releases a broken

JohanMabille:
The alpha candidates of micromamba are now listed in `https://api.anaconda.org/release/conda-forge/micromamba/latest`, which is mirrored and used in production environments. Moving the alpha candidates to the broken channel should fix this, until we find a way to filter alpha releases from `latest` mirrors.


xhochy:
These packages are on a label. We cannot mark things that aren't on main as broken.

JohanMabille:
How can I remove them? Ot does anaconda provide an API for getting the latest "stable" release of a package?

PR #904, opened 2024-01-05
Add core members to heroku and anaconda

isuruf:

## Checklist:

### DIFF
--- a/conda_forge_admin_requests/core.py
+++ b/conda_forge_admin_requests/core.py
@@ -4,6 +4,18 @@ import tempfile
 import requests
 import copy
 from conda_smithy.github import Github
+import csv
+
+
+def _get_core(url):
+    res = requests.get(url)
+    data = csv.reader(res.content.decode("utf-8").splitlines(), delimiter=',')
+    return set([row[0] for row in list(data)[1:]])
+
+
+def get_core():
+    URL_BASE = "https://raw.githubusercontent.com/conda-forge/conda-forge.github.io/main/src/"
+    return _get_core(f"{URL_BASE}/core.csv") | _get_core(f"{URL_BASE}/emeritus.csv")
 
 
 def check(request):
@@ -14,12 +26,17 @@ def check(request):
     gh_handle = request["github"]
     # TODO: check that the PR was submitted by the user
 
+    pr_author = os.environ["GITHUB_PR_AUTHOR"]
+    assert pr_author == gh_handle
+
+    core = get_core()
+    assert pr_author in core
+
 
 def run(request):
     action = request.pop("action")
     assert action == "core"
     assert "github" in request
-    check(request)
 
     gh_handle = request.pop("github")
     gh = Github(os.environ['GITHUB_TOKEN'])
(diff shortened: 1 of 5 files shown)


isuruf:
@conda-forge/core, thoughts on this. I might add other things like `quay.io`, `dockerhub` etc if this is okay. There might be security implications I haven't thought about.

beckermr:
Definitely a good idea. I'd be happier if we moved this to a private a repo so we don't accidentally merge a random PR without thinking. We could use one of the ones we already have.

jakirkham:
Thanks Isuru! 🙏

Agree this is a good idea

Matt's point also sounds reasonable

jakirkham:
Thinking about this more. Would it make sense to add this info to [`core.csv `]( https://github.com/conda-forge/conda-forge.github.io/blob/main/src/core.csv ) and leverage that to update access?

beckermr:
I don't follow that last comment @jakirkham. However I am ok with the checks just added by @isuruf. 

We should add one more check that the only files modified are adding the yaml and then I'd be ok merging. This way we'll have a backup if someone attempts to modify the scripts to remove the check for core members.

jakirkham:
Meaning that we would include our DockerHub, Quay, and Heroku handles in `core.csv` file and adding them there would have the effect of adding access to the different services

conda_forge_admin_requests/core.py
```diff
@@ -0,0 +1,71 @@
...
+import os
+import tempfile
+import requests
+import copy
+from conda_smithy.github import Github
+import csv
+
+
+def _get_core(url):
+    res = requests.get(url)
```
jaimergp, line 11:
```suggestion
    res = requests.get(url)
```
```suggestion
    res = requests.get(url)
    res.raise_for_status()
```

conda_forge_admin_requests/core.py
```diff
@@ -0,0 +1,71 @@
...
+    data = csv.reader(res.content.decode("utf-8").splitlines(), delimiter=',')
```
jaimergp, line 12: This can be [`.text`](https://requests.readthedocs.io/en/latest/user/quickstart/#response-content), I think?

```suggestion
    data = csv.reader(res.text.splitlines(), delimiter=',')
```

conda_forge_admin_requests/core.py
```diff
@@ -0,0 +1,71 @@
...
+    return set([row[0] for row in list(data)[1:]])
+
+
+def get_core():
+    URL_BASE = "https://raw.githubusercontent.com/conda-forge/conda-forge.github.io/main/src/"
+    return _get_core(f"{URL_BASE}/core.csv") | _get_core(f"{URL_BASE}/emeritus.csv")
+
+
+def check(request):
```
jaimergp, line 21: Should we run `check()` as part of `run()` too?

isuruf:
Going to use this in a private repo. See https://github.com/conda-forge/admin-requests/pull/906
Thanks everyone for the feedback.

PR #911, opened 2024-01-15
Mark azure-core-cpp 1.11 broken on linux

jorisvandenbossche:
See https://github.com/conda-forge/tiledb-feedstock/issues/228#issuecomment-1890919858 for context. This latest release breaks tiledb (which doesn't have a strict pin on the azure-core-cpp version, just <2), and in that way also breaks eg libgdal.

In our CI where we noticed this, it's only on Linux that we see the crash, not on Windows or Mac. Therefore I only marked the linux packages as broken. We only test linux-64, but I marked aarch64 and pplc as well (assuming those would have the same issue?)

cc @conda-forge/azure-core-cpp 

## Checklist:

### DIFF
--- a/requests/azure-core-cpp-1-11.yml
+++ b/requests/azure-core-cpp-1-11.yml
@@ -1,0 +1,11 @@
+action: broken
+packages:
+  - linux-64/azure-core-cpp-1.11.0-h22d5011_0.conda
+  - linux-64/azure-core-cpp-1.11.0-hd64b3e6_0.conda
+  - linux-64/azure-core-cpp-1.11.0-h91d86a7_0.conda
+  - linux-aarch64/azure-core-cpp-1.11.0-haf42898_0.conda
+  - linux-aarch64/azure-core-cpp-1.11.0-hc0db439_0.conda
+  - linux-aarch64/azure-core-cpp-1.11.0-hcd87347_0.conda
+  - linux-ppc64le/azure-core-cpp-1.11.0-h0c291d7_0.conda
+  - linux-ppc64le/azure-core-cpp-1.11.0-he7d803c_0.conda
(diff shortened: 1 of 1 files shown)


jjerphan (approved):
LGTM.

Thank you for having taken care of it, @jorisvandenbossche. :slightly_smiling_face:

jdblischak:
I appreciate that speed was desired to fix this, but I think this would have been better solved with a repodata patch. There is nothing inherently wrong with azure-core-cpp 1.11.0 binaries. The problem is that it broke `tiledb`, so we need to patch its metadata to include an upper limit of `<1.11`. The azure-core-cpp recipe uses "run exports", so if any conda binaries were built against azure-core-cpp 1.11.0 over the weekend, they are now unusable (because it uses the default of `min_pin="x.x.x.x"`).

Since there is nothing wrong with the 1.11.0 binaries themselves, I propose the following plan:

- [ ] Patch the `tiledb` repodata to include `azure-core-cpp < 1.11`
- [ ] Mark the azure-core-cpp binaries as unbroken
- [ ] Update the `tiledb` recipe to specify `azure-core-cpp < 1.11`

Longer term, I think we should update the run exports of azure-core-cpp to be `max_pin="x.x"`. We trusted the upstream developers that minor patches would be backwards compatible, but that clearly isn't the case. Had the setting been `max_pin="x.x"`, then the new azure-core-cpp 1.11 would not have been installed with `tiledb` in the first place. We should also consider this for the other Azure C++ SDK library recipes.

cc @teo-tsirpanis

jdblischak:
Sorry, I missed the discussion in https://github.com/conda-forge/tiledb-feedstock/issues/228 where it was decided to mark them as broken

PR #917, opened 2024-01-16
Request to remove broken build 0 of xarray-2023.10.0

julian-belina:
Hello,

I would like to report the build 0  of xarray-2023.10.0 as broken. However I am not sure if a meta data patch or the deletion of the package is the correct proceedure. There are two builds of the version (xarray-2023.10.0) which only differ in there dependency meta data. (numpy >=1.21 and numpy >=1.22 respectively)
 
https://github.com/conda-forge/xarray-feedstock/commit/7047d15b508b860bd71eb2d231d921e5c699a87b

Should the meta data still be patched even though build 0 and 1 are exactly the same then when patched?

All the best,

Julian

Thank you for making an admin request on this repo. We strive to make a decision
on these requests within 24 hours. 

Please use the text below to add context about this PR, especially if:
- You want to mark packages as broken
- You want to archive a feedstock
- You want to request access to opt-in CI resources

Cheers and thank you for contributing to conda-forge!
-->

## Checklist:

### DIFF
--- a/requests/xarray_2023_10_0_build_0-broken.yml
+++ b/requests/xarray_2023_10_0_build_0-broken.yml
@@ -1,0 +1,3 @@
+action: broken
+packages:
+  - noarch/xarray-2023.10.0-pyhd8ed1ab_0.conda


julian-belina:
@cona-forge/xarray , could you give your input on the broken build?

hmaarrfk:
i think it is better to repodata patch.

hmaarrfk:
if 2023.10 was released today, i would have just merged this and left the repodata patch for an other day, but the fact that it was live for like 2 months (assuming it was released on conda-forge in 2023.10) makes it less urge to mark as broken.

jhamman:
> @conda-forge/xarray , could you give your input on the broken build?

Not sure what the correct thing to do here is. If there is a conda-forge process for updating (or removing) broken builds, I'd be happy for someone to push on that. 

However, Xarray has had 3 releases since the one in question so if fixing this is going to eat up a ton of time, I'd consider finding another way to work around the problem.

hmaarrfk:
done in https://github.com/conda-forge/conda-forge-repodata-patches-feedstock/pull/635

thanks for raising this issue!!!

PR #927, opened 2024-01-25
mark older builds of kim-api-2.3.0 as broken

ilia-nikiforov-umn:
We have recently added the ability to build models to the kim-api conda-forge package. This is not a new functionality of the KIM API, it just was not present in the conda-forge package. This involved adding compilers (e.g. {{ compiler('c') }}, etc.) and build tools as run requirements, as well as a modification to the build script in order to set paths to compilers through CMake (so this is not just a metadata deficiency).

Because older builds of kim-api-2.3.0 don't have these dependencies, the solver will preferentially install them if any conflicting versions of the compilers are present in the environment. This is confusing to users, who expect the fixed model-building functionality to be present in their environment, but it is not because they accidentally obtained an older build without any warning. Because of this, we think it's appropriate to mark the older builds as broken (after all, they are missing functionality they're supposed to have). I think just marking the 2.3.0 builds as broken is sufficient, this way we can unambiguously instruct users to install kim-api=2.3.0

I've already discussed this with [kim-api-feedstock](https://github.com/conda-forge/kim-api-feedstock) developers @jan-janssen @ellio167

## Checklist:

### DIFF
--- a/requests/kim-api-2.3.0-broken.yml
+++ b/requests/kim-api-2.3.0-broken.yml
@@ -1,0 +1,9 @@
+action: broken
+packages:
+- osx-arm64/kim-api-2.3.0-h08610c0_0.conda
+- osx-64/kim-api-2.3.0-hb20c0b4_0.conda
+- linux-64/kim-api-2.3.0-h94b5c5c_0.conda
+- linux-aarch64/kim-api-2.3.0-h5a3f854_0.conda
+- osx-64/kim-api-2.3.0-h32d3758_0.tar.bz2
+- osx-arm64/kim-api-2.3.0-h9c9afe6_0.tar.bz2
+- linux-64/kim-api-2.3.0-h0911212_0.tar.bz2


jan-janssen (approved):
Looks good to me

ocefpaf:
We could probably do a repodata patch in that case. Do you want to try it? If I understood the problem correctly the parch should add the compilers package as a dependency on these builds, right?

ilia-nikiforov-umn:
> We could probably do a repodata patch in that case. Do you want to try it? If I understood the problem correctly the parch should add the compilers package as a dependency on these builds, right?

Hi Filipe,

Here is the pull request that added the missing functionality:

https://github.com/conda-forge/kim-api-feedstock/pull/18/files

As you can see, it's not just a matter of dependencies, the build script was changed as well.

ocefpaf:
> As you can see, it's not just a matter of dependencies, the build script was changed as well.

Then we cannot patch. Merging this one. Thanks!

PR #945, opened 2024-02-22
PRs and CPU runners for cutlass

RaulPPelaez:
Cutlass takes about 24 hours to compile in the current runners. I am trying to make it run on the openstack ones.

## Checklist:

### DIFF
--- a/requests/cutlass.yaml
+++ b/requests/cutlass.yaml
@@ -2,8 +2,7 @@ action: cirun
 feedstocks:
   - cutlass
 resources:
-  - cirun-openstack-gpu-large
-  - cirun-openstack-cpu-large
+  - cirun-openstack-cpu-4xlarge
 pull_request: true
 revoke: false
 send_pr: true
--- a/requests/cutlass.yaml
+++ b/requests/cutlass.yaml
@@ -2,7 +2,7 @@ action: cirun
 feedstocks:
   - cutlass
 resources:
-  - cirun-openstack-cpu-4xlarge
+  - cirun-openstack-cpu-large
 pull_request: true
 revoke: false
 send_pr: true
--- a/requests/cutlass.yaml
+++ b/requests/cutlass.yaml
@@ -1,0 +1,8 @@
+action: cirun
+feedstocks:
+  - cutlass
+resources:
+  - cirun-openstack-gpu-large
+  - cirun-openstack-cpu-large
+pull_request: true
+revoke: false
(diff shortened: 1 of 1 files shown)


requests/cutlass.yaml
```diff
@@ -0,0 +1,8 @@
+action: cirun
+feedstocks:
+  - cutlass
+resources:
+  - cirun-openstack-gpu-large
+  - cirun-openstack-cpu-large
+pull_request: true
+revoke: false
```
h-vetinari, line 8:
```suggestion
revoke: false
send_pr: true
```

requests/cutlass.yaml
```diff
@@ -0,0 +1,8 @@
+action: cirun
+feedstocks:
+  - cutlass
+resources:
+  - cirun-openstack-gpu-large
+  - cirun-openstack-cpu-large
```
h-vetinari, line 6: I don't actually think we'll need the CPU version, as cutlass only has GPU builds.

h-vetinari (approved):
So this is basically a copy of #881 but for cutlass, which is a good start; however, pytorch was the very first, and there were a few more iterations since then.

It's better to take the [example](https://github.com/conda-forge/admin-requests/blob/main/examples/example-open-gpu-server.yml) as a reference.

requests/cutlass.yaml
RaulPPelaez: The GPU I do not think its actually needed. We are going to compile but not test.
We could test on the V100 available in the self-hosted runner, but I believe it is a waste to hold the GPU during 24h just for that...

RaulPPelaez: In that sense cutlass would benefit from ci_4xlarge, which has 8 cores and a lot of RAM

h-vetinari: > We could test on the V100 available in the self-hosted runner, but I believe it is a waste to hold the GPU during 24h just for that...

The openstack runner is _much_ more powerful than the generic azure agents. I think if the pytorch megabuild runs through with `cirun-openstack-gpu-large`, it's unlikely we'll need something bigger for cutlass. But on that I'll defer to @jaimergp and @isuruf

jaimergp: Let's try with `large` first and if it proves to be insufficient, we can check xlarge. Agree that there's no GPU required in this case.

requests/cutlass.yaml
```diff
@@ -0,0 +1,8 @@
+action: cirun
+feedstocks:
+  - cutlass
+resources:
+  - cirun-openstack-cpu-4xlarge
```
jaimergp, line 5:
```suggestion
  - cirun-openstack-cpu-large
```

RaulPPelaez:
Thanks guys, always a pleasure.

PR #948, opened 2024-02-23
enable tensorflow to use cirun

h-vetinari:
Tensorflow has long been stuck in CFEP-03 land, let's try to get it out of there. ;-)

CC @jaimergp @isuruf @hmaarrfk @xhochy

### DIFF
--- a/requests/tensorflow.yaml
+++ b/requests/tensorflow.yaml
@@ -1,0 +1,8 @@
+action: cirun
+feedstocks:
+  - tensorflow
+resources:
+  - cirun-openstack-cpu-large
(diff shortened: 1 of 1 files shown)


hmaarrfk:
Just to confirm, will you be trying to fit it into the cirun constraints?

h-vetinari:
> Just to confirm, will you be trying to fit it into the cirun constraints?

I'm expecting we can find a combination of [server size](https://github.com/Quansight/open-gpu-server#available-runners) & timeout that'll be able to build the current recipe in its entirety. Do you have doubts that this will be possible?

PR #949, opened 2024-02-23
enable mongodb to use cirun

h-vetinari:
Current mongodb v7 manages to compile about half the objects before running into the 6h timecap ([example](https://dev.azure.com/conda-forge/feedstock-builds/_build/results?buildId=882735&view=results); around `[2385+2/4544]` on linux-64, `[2398+2/4489]` on win-64, etc.) - completely hopeless to get it over the line.

CC @jaimergp

### DIFF
--- a/requests/mongodb.yaml
+++ b/requests/mongodb.yaml
@@ -1,0 +1,8 @@
+action: cirun
+feedstocks:
+  - mongodb
+resources:
+  - cirun-openstack-cpu-large
+pull_request: true
+revoke: false
+send_pr: true


jaimergp:
The server can help with Linux builds but we can't do much for Windows, though. I wonder if it's just that mongodb has increased in complexity over time, or whether there's something off in the build.

Also, tagging @conda-forge/mongodb to follow protocol (but I guess we won't hear back, the feedstock has been inactive for a while).

h-vetinari:
> The server can help with Linux builds but we can't do much for Windows, though.

We could conceivably cross-compile from windows from linux. We have [`linux-64/clang_win-64`](https://github.com/conda-forge/clang-win-activation-feedstock/), but I've never tried to cross-compile across OSes. Perhaps @isuruf or @xhochy have thoughts.

> Also, tagging @conda-forge/mongodb to follow protocol (but I guess we won't hear back, the feedstock has been inactive for a while).

I did a big [push](https://github.com/conda-forge/mongodb-feedstock/pull/78) late last year to revive it, but didn't get all the way up-to-date due to the timeouts. Now that we have the cirun server, I thought I'd give it another shot.

xhochy:
> We could conceivably cross-compile from windows from linux. We have [linux-64/clang_win-64](https://github.com/conda-forge/clang-win-activation-feedstock/), but I've never tried to cross-compile across OSes. Perhaps @isuruf or @xhochy have thoughts.

The only feedstock where we do this is https://github.com/conda-forge/polars-feedstock There is one major constraint though: You cannot have a host environment in that situation. There is a major conda-build bug with prefix replacement. I would rather wait until we have `rattler-build` in conda-forge than tackling this.

h-vetinari:
> There is one major constraint though: You cannot have a host environment in that situation. There is a major conda-build bug with prefix replacement.

It seems Isuru recently tackled (at least part of) this: https://github.com/conda/conda/issues/13687 🙏

PR #967, opened 2024-03-19
token refresh for batchspawner

bollwyvl:
This requests a token refresh for https://github.com/conda-forge/batchspawner-feedstock/issues/6


beckermr (changes requested):
Please use the new yaml format for the requests.

bollwyvl:
I wrote a [thing](https://github.com/deathbeds/jupyak/tree/main/js/demo) that builds clickable links for single-file pull requests of json/toml/yaml from JSON schema... 

https://deathbeds.github.io/jupyak/new.html

... maybe could apply that here.

beckermr:
Oh cool! Def make an issue with the idea.

bollwyvl:
:point_up: #968

beckermr:
Ohhhh woops I missed that! I bet we could do this as part of the new docs repo maybe? I'm not sure how api access would work to generate a pr though. I don't want to handle credentials. 

Maybe an admin command could be interesting and easier from a security perspective?

Cc @jaimergp

PR #971, opened 2024-03-28
Mark graphite2-1.3.13-*_1002 Windows package as broken

traversaro:
`conda create -n ffmpeg ffmpeg` On Windows currently results in a broken package, and similarly `import cv2` crashes, as harfbuzz can't find `graphite2.dll`. The package contain a file called `libgraphite2.dll` instead, as I guess it was compiled with mingw. See https://github.com/conda-forge/graphite2-feedstock/issues/14 for more details, however I think it is a good idea to just mark it as broken to fix all the library that depends on harfbuzz.

cc @conda-forge/ffmpeg @conda-forge/opencv @conda-forge/harfbuzz @conda-forge/graphite2 

## Checklist:

### DIFF
--- a/requests/graphite2-broken.yml
+++ b/requests/graphite2-broken.yml
@@ -1,0 +1,3 @@
+action: broken
+packages:
+- win-64/graphite2-1.3.13-h57928b3_1002.conda


traversaro:
Thanks @ocefpaf !

PR #1032, opened 2024-07-17
Mark broken

Sparks29032:

## Checklist:

Want to mark `bg_mpl_stylesheets` as broken as the correct package name is `bg-mpl-stylesheets`. We do not want to maintain the misnamed `bg_mpl_stylesheets`, so we will mark as broken to ensure users do not download this unmaintained version in the future.

ping @conda-forge/bg-mpl-stylesheets

### DIFF
--- a/requests/bg_mpl_stylesheets-broken.yml
+++ b/requests/bg_mpl_stylesheets-broken.yml
@@ -1,0 +1,4 @@
+action: broken
+packages:
+  - noarch/bg_mpl_stylesheets-0.3.0-pyhd8ed1ab_1.conda
+  - noarch/bg_mpl_stylesheets-0.3.0-pyhd8ed1ab_0.conda


jaimergp:
What we usually do here is to add an alias package so `bg_mpl_stylesheets` depends on `bg-mpl-stylesheets` and nothing is broken. See these two examples:

- https://github.com/conda-forge/importlib_metadata-feedstock/blob/main/recipe/meta.yaml
- https://github.com/conda-forge/typing_extensions-feedstock/blob/main/recipe/meta.yaml

This would result in less confusion for downstream users who had the underscore variant in their environment.yml etc.

Sparks29032:
Thanks, we will discuss this before deciding whether to close this PR!

sbillinge:
I like the solution from @jaimergp if we can do it.

carterbox:
Only these two artifacts were ever published, then the name was reverted within a day. All other versions 0.1.0 to 0.6.1 use - instead of _.

https://github.com/conda-forge/bg-mpl-stylesheets-feedstock/commit/48621894d024fa0ae706cb639ea9f771ccdd0720

https://github.com/conda-forge/bg-mpl-stylesheets-feedstock/pull/12/changes

sbillinge:
thanks @carterbox !

PR #1033, opened 2024-07-18
Mark rattler-build-compat-broken as broken

isuruf:
See https://github.com/conda-forge/rattler-build-conda-compat-feedstock/issues/11

We can't patch repodata since building that requires building a feedstock and all feedstocks are broken right now.

cc @conda-forge/rattler-build-conda-compat

### DIFF
--- a/requests/rattler-build-compat-broken.yml
+++ b/requests/rattler-build-compat-broken.yml
@@ -1,0 +1,3 @@
+action: broken
+packages:
+- noarch/rattler-build-conda-compat-0.1.2-pyhd8ed1ab_0.conda


isuruf:
@conda-forge/core, can I get a quick review?

isuruf:
Going to merge this for now to unblock all the feedstocks. See https://github.com/conda-forge/status/issues/183

beckermr:
I don't follow sorry. 

FWIW, putting a broken label on anaconda.org and then waiting for a CDN sync won't actually move the package to broken. You probably know this already. 

The best way to procede would be to hack the repodata-pacthes feedstock build scripts to use an older version of smithy. Maybe you did that already?

jaimergp:
Can't we skip the broken rattler-build-conda-compat package in the repodata-patches feedstock via conda-forge.yml? With something like `conda_forge_ci_setup: ["rattler-build-conda-compat!=0.1.2"]`?

beckermr:
Sure that'd work too. Either way we need to do something more beyond applying a label and waiting.

jaimergp:
Preparing a PR

jaimergp:
https://github.com/conda-forge/conda-forge-repodata-patches-feedstock/pull/801

isuruf:
Instead of having both `main` and `broken`, it has `broken` only which means it will not get synced to the CDN.

beckermr:
Ohhhhhh. Thanks @isuruf. That's correct. I thought you had only added the broken label.

PR #1034, opened 2024-07-19
mark the RDKit MacOS 2024.3.4 build 1 packages as broken

greglandrum:
## Description of the problem

An update to the conda-forge version of `libcxx` triggered a bug in some very basic functionality within the rdkit packages on MacOS (intel and arm64), rendering those packages unuseable for many people.

This PR will remove build 1 of these MacOS packages, leaving build 0, which used the older version of `libcxx` and so doesn't trigger the bug.

More context and a link to the upstream fixes here:
https://github.com/conda-forge/rdkit-feedstock/issues/162

This will, unfortunately, remove the numpy 2.0 compatible RDKit builds for MacOS.
We're also doing an upstream release today so that we can get an up-to-date RDKit version (with numpy 2.0 support) out there for all systems.

## Checklist:

### DIFF
--- a/requests/rdkit-broken.yml
+++ b/requests/rdkit-broken.yml
@@ -1,0 +1,12 @@
+action: broken
+packages:
+- osx-arm64/rdkit-2024.03.4-py39h1e940ea_1.conda
+- osx-arm64/rdkit-2024.03.4-py38h85cd035_1.conda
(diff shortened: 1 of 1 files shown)


greglandrum:
Thanks @jaimergp

PR #1040, opened 2024-07-26
Request large CPU/GPU runners for flash-attn

weiji14:
Applying for `cirun-openstack-gpu-large` and `cirun-openstack-cpu-large` for building flash-attn. Builds were originally using `TORCH_CUDA_ARCH_LIST=8.0+PTX` only to try to keep under 6 hour limit at https://github.com/conda-forge/staged-recipes/pull/26239, but they recently started to timeout at https://github.com/conda-forge/flash-attn-feedstock/pull/10.

Using a bigger instance will also allow us to build for `TORCH_CUDA_ARCH_LIST=8.0;8.6;8.9;9.0+PTX`, xref https://github.com/conda-forge/flash-attn-feedstock/issues/4

## Checklist:

### DIFF
--- a/requests/flash-attn-open-gpu-server.yml
+++ b/requests/flash-attn-open-gpu-server.yml
@@ -1,0 +1,8 @@
+action: cirun
+feedstocks:
+  - flash-attn
+resources:
+  - cirun-openstack-gpu-large
+pull_request: true
+revoke: false
+send_pr: true
--- a/requests/flash-attn-open-gpu-server.yml
+++ b/requests/flash-attn-open-gpu-server.yml
@@ -3,6 +3,7 @@ feedstocks:
   - flash-attn
 resources:
   - cirun-openstack-gpu-large
+  - cirun-openstack-cpu-large
 pull_request: true
 revoke: false
 send_pr: true


jaimergp:
Would you like access to the CPU runners too? This way you can mix e.g. one GPU job with three CPU ones, because there re only 4 GPU slots available.

carterbox:
Yes. We can use CPU runners because we don't execute any tests presently. I'm not sure how to specify whether the job is sent to CPU or GPU runner though.

jaimergp:
For now add it to the request here. Then we will adjust the conda-build-config file.

requests/flash-attn-open-gpu-server.yml
```diff
@@ -0,0 +1,8 @@
+action: cirun
+feedstocks:
+  - flash-attn
+resources:
+  - cirun-openstack-gpu-large
```
carterbox, line 5:
```suggestion
resources:
  - cirun-openstack-gpu-large
  - cirun-openstack-cpu-large
```

jaimergp (approved):

PR #1041, opened 2024-07-31
Request to mark pip 24.1.2 as broken.

tribeiro:

## Checklist:

I opened an issue in the package feedstock about the issue we are experiencing with this version of the package (conda-forge/pip-feedstock#124). The change was actually intentional and we are now discussing whether they are going to keep the change or not. I am opening the PR now to make sure it is ready to go if they decide to rollback the change.

### DIFF
--- a/requests/pip-broken.yml
+++ b/requests/pip-broken.yml
@@ -1,0 +1,3 @@
+action: broken
+packages:
+  - noarch/pip-24.1.2-pyhd8ed1ab_0.conda


tribeiro:
I think we arrived at a consensus to undo the change for the time being. So I am opening this up again to mark the package as broken.

jakirkham:
@conda-forge/core , do we want to...?

1. Mark these old packages as `broken` (done in this PR)
2. Repodata patch the old packages to depend on `wheel` & `setuptools`
3. Leave as-is (since newer packages don't have this issue)

beckermr:
Marking broken is easiest.

tsibley:
@beckermr wrote:

> Marking broken is easiest.

FWIW, marking as broken broke our reproducible environments which incorporated this package build. I think patching the repo data (@jakirkham's option 2) would have avoided that. The guidelines from the PR description support that too:

> We prefer to patch the repo data (see [here](https://github.com/conda-forge/conda-forge-repodata-patches-feedstock)) instead of marking packages as broken. This alternative workflow makes environments more reproducible.

It would have been nicer for us if the choice made was to patch the repo data instead of marking broken, even if marking broken was easier.

In any case, thank you for the work y'all do to keep the Conda and Pip/Python ecosystems up-to-date and moving forward! I know it's a job that's not easy and often thankless. :pray:

beckermr:
Oh no! We can undo this. Reproducible envs made with tools like `pixi`, explicit conda list exports, and `pixi` do not have that issue. My appoligies and I will fix this later.

beckermr:
I'm curious @tsibley about what workflow you use that this broke. If you can send me some links, that'd be helpful. 

FWIW, environments exported via `conda env export > environment.yml` are inherently not reproducible. The reason is that envs made this way have to be resolved using the solver. That process depends on the repodata.json from anaconda.org which is updated every ~10 minutes or so. Thus this process is inherently not reproducible. If you instead use conda-lock, pixi or similar, the PR would leave things unaffected.

tsibley:
@beckermr Not a big deal, but I wanted to note the breakage. What we're doing is creating meta-packages that strictly pin their dependencies at build/render time, akin to [`pin_depends: strict`](https://docs.conda.io/projects/conda-build/en/stable/resources/define-metadata.html#pin-runtime-dependencies) I think but implemented differently due to performance constraints with `conda` at the time (I suspect better now!) that forced us to use `boa`. We then create new envs with just that meta-package installed, and yes, that uses the solver. But using meta-packages lets us easily version and release these meta-packages and upgrade envs without downloading packages that haven't changed. Most of the time the solver isn't an issue, but yes, this approach can be broken by upstream actions that rewrite history like yanking packages from the index or repo data patches that substantially alter them to cause conflicts (we've seen this before). You can see our build repo at https://github.com/nextstrain/conda-base.

beckermr:
Right OK. I'd strongly suggest moving away from this workflow as we at conda-forge cannot promise it won't get broken by our actions. For you all, [conda-lock](https://github.com/conda/conda-lock) would be a suitable, but somewhat different solution.

tsibley:
Nod. Understood. I figured that would be your suggestion. :-) We'll have to consider the tradeoffs of publishing lock files ourselves instead of meta-packages via Anaconda. (We already use Docker images we produce when we do really want stricter reproducibility.)

beckermr:
xref: https://github.com/conda-forge/conda-forge-repodata-patches-feedstock/pull/810
xref: https://github.com/conda-forge/admin-requests/pull/1042

jakirkham:
Thanks Matt! 🙏

PR #1046, opened 2024-08-21
Do not use `shell=True` in subprocesses

jaimergp:
As per recommendations at https://docs.python.org/3/library/subprocess.html#security-considerations.

### DIFF
--- a/conda_forge_admin_requests/access_control.py
+++ b/conda_forge_admin_requests/access_control.py
@@ -74,7 +74,9 @@ def send_pr_cirun(
 
     git_cmds = [
         ["git", "add", "recipe/conda_build_config.yaml", "conda-forge.yml"],
-        ["git", "remote", "add", user.login, f"https://x-access-token:${{GITHUB_TOKEN}}@github.com/{user.login}/{feedstock}.git"],
+        ["git", "remote", "add", user.login, 
+         f"https://x-access-token:{os.environ['GITHUB_TOKEN']}@github.com/{user.login}/{feedstock}.git"
+        ],
         ["git", "commit", "-m", f"Enable {resource_str} using Cirun", "--author", f"{user.name} <{user.email}>"],
         ["conda-smithy", "rerender", "-c", "auto", "--no-check-uptodate"],
         ["git", "push", user.login, f"HEAD:{base_branch}"],
@@ -137,7 +139,7 @@ def _process_request_for_feedstock(
 
         owner_info = ["--organization", GH_ORG]
         token_repo = (
-            'https://x-access-token:${GITHUB_TOKEN}@github.com/'
+            f'https://x-access-token:{os.environ['GITHUB_TOKEN']}@github.com/'
             f'{GH_ORG}/feedstock-tokens'
         )
 
--- a/conda_forge_admin_requests/access_control.py
+++ b/conda_forge_admin_requests/access_control.py
@@ -73,16 +73,15 @@ def send_pr_cirun(
     resource_str = ", ".join(resources)
 
     git_cmds = [
-        "git add recipe/conda_build_config.yaml conda-forge.yml",
-        f"git remote add {user.login} https://x-access-token:${{GITHUB_TOKEN}}@github.com/"
-        f"{user.login}/{feedstock}.git",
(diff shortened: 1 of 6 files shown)


conda_forge_admin_requests/mark_broken.py
```diff
@@ -89,33 +90,46 @@ def run(request):
...
+                    "https://github.com/conda-forge/conda-forge-repodata-patches-feedstock.git",
+                ],
                 cwd=tmpdir,
-                shell=True,
             )
 
-            subprocess.check_call(
-                "git remote set-url --push origin "
+            origin_url = (
                 "https://x-access-token:${GITHUB_TOKEN}@github.com/conda-forge/"
```
jaimergp, line 102: Not sure if `git` recognizes this env var syntax on its own or whether we were relying on shell expansion to populate this.

beckermr: We are relying on shell expansion here.

conda_forge_admin_requests/access_control.py
```diff
@@ -73,16 +73,15 @@ def send_pr_cirun(
...
 
     git_cmds = [
-        "git add recipe/conda_build_config.yaml conda-forge.yml",
-        f"git remote add {user.login} https://x-access-token:${{GITHUB_TOKEN}}@github.com/"
-        f"{user.login}/{feedstock}.git",
-        f"git commit -m 'Enable {resource_str} using Cirun' --author '{user.name} <{user.email}>'",
-        "conda-smithy rerender -c auto --no-check-uptodate",
-        f"git push {user.login} HEAD:{base_branch}",
+        ["git", "add", "recipe/conda_build_config.yaml", "conda-forge.yml"],
+        ["git", "remote", "add", user.login, f"https://x-access-token:${{GITHUB_TOKEN}}@github.com/{user.login}/{feedstock}.git"],
```
beckermr, line 77: This is relying on shell expansion.

update_repodata_patches.py
```diff
@@ -64,29 +61,39 @@ def update_repodata_patches(dry_run):
...
+                "git",
+                "clone",
+                "https://github.com/conda-forge/conda-forge-repodata-patches-feedstock.git",
+            ],
             cwd=tmpdir,
-            shell=True,
         )
 
+        origin_url = (
+            "https://x-access-token:${GITHUB_TOKEN}@github.com/"
```
beckermr, line 73: More shell expansion.

beckermr (changes requested):
I flagged the shell expansion locations for reference.

jaimergp:
Turned them into `os.environ['GITHUB_TOKEN']` calls.

conda_forge_admin_requests/access_control.py
```diff
@@ -138,7 +139,7 @@ def _process_request_for_feedstock(
 
         owner_info = ["--organization", GH_ORG]
         token_repo = (
-            'https://x-access-token:${GITHUB_TOKEN}@github.com/'
+            f'https://x-access-token:{os.environ['GITHUB_TOKEN']}@github.com/'
             f'{GH_ORG}/feedstock-tokens'
```
beckermr, line 143:
```suggestion
            f"https://x-access-token:{os.environ['GITHUB_TOKEN']}@github.com/"
            f"{GH_ORG}/feedstock-tokens"
```

beckermr:
The github CI should do the scrubbing for us.

jaimergp:
If it's a secret Github Actions will recognize it in the stdout and mask it anyway. The only concern might be that the stdout is somehow mangled (line wrapped, etc) and that the secret is split in half or something, but I don't see why that would happen. 

The tokens here are short-lived anyway (they just last as long as GHA session does, right?):

> Before each job begins, GitHub fetches an installation access token for the job. The GITHUB_TOKEN expires when a job finishes or after a maximum of 24 hours.

So even if they leaked, the damage is really contained.

jaimergp:
@isuruf, are your concerns alleviated or would you prefer if I add a custom [`sys.excepthook`](https://docs.python.org/3/library/sys.html#sys.excepthook) to filter the tracebacks ourselves?

isuruf (approved):

jaimergp:
@beckermr, any concerns?

beckermr (approved):

PR #1047, opened 2024-08-21
Add workflow for CFEP3 copies

jaimergp:

## Checklist:

For provenance and transparency, this workflow will allow to process manual CFEP3 copies without any core member needing a token locally. The review process must still happen, though.


beckermr (changes requested):
We should have folks specify a SHA256 hash in the yaml request that we check against the value from anaconda.org before copying.

conda_forge_admin_requests/cfep3_copy.py
```diff
@@ -7,50 +7,65 @@
...
+        raise ValueError(f"Package '{package}' at {channel} does not seem to exist")
+    
+    # Check SHA256
+    r = requests.get(
+        f"https://api.anaconda.org/dist/{channel}/{pkg_name}/{version}/{subdir}/{artifact}",
+        timeout=10,
+    )
+    r.raise_for_status()
+    api_sha256 = r.json()["sha256"]
+    if sha256 != api_sha256:
```
beckermr, line 38:
```suggestion
    if not hmac.compare_digest(sha256, api_sha256):
```

You should always use this function when comparing digests to prevent timing attacks: https://docs.python.org/3/library/hmac.html#hmac.compare_digest

examples/example-cfep-3.yml
```diff
@@ -1,5 +1,5 @@
 action: cfep3_copy
-anaconda_org_packages: # list of owner[@label=main]/pkg-name/version/subdir/filename.extension
- - jaimergp/conda-standalone/24.7.1/osx-arm64/conda-standalone-24.7.1-hce30654_0.conda
-from_anaconda_org_label: main # optional, default=main
+anaconda_org_packages:
+  # dict of (sha256, channel[/label/<label>]/subdir/filename.extension)
+  cd11f1f0fbe8c203212d002ca4e01795d0b1c886ce47d8b08e6bca3366a7c8e6: jaimergp/osx-arm64/conda-standalone-24.7.1-hce30654_0.conda
```
beckermr, line 4: Some how the ordering here feels backwards. Maybe map package name to sha, or does that have other issues?

beckermr (changes requested):

examples/example-cfep-3.yml
jaimergp: The sha256 is guaranteed to not contain a `:`, but I'm not sure if the label can have one. If that's the case, then the user would be forced to quote the key to prevent YAML from considering the `:` as a key, value separator. But tbh it's such an edge case that it doesn't really matter.
We can also do something like:

```yaml
- package: jaimergp/label/conda-standalone-24.7.1/osx-arm64/conda-standalone-24.7.1-hce30654_0.conda
  sha256: cd11f1f0fbe8c203212d002ca4e01795d0b1c886ce47d8b08e6bca3366a7c8e6
```

if you prefer it.

beckermr: Ahhhhh that makes sense. In yaml you can add quotes around the key for these edge cases. 

I like the list of entries better so let's go with that.

jaimergp: Done!

examples/example-cfep-3.yml
```diff
@@ -1,5 +1,6 @@
 action: cfep3_copy
 anaconda_org_packages:
-  # dict of (sha256, channel[/label/<label>]/subdir/filename.extension)
-  cd11f1f0fbe8c203212d002ca4e01795d0b1c886ce47d8b08e6bca3366a7c8e6: jaimergp/osx-arm64/conda-standalone-24.7.1-hce30654_0.conda
+  # list of {package: channel[/label/<label>]/subdir/filename.extension, sha256: SHA256}
```
beckermr, line 3: Doesn't conda-build use a `::` syntax for this? Maybe we should use that?

beckermr: By this I mean prepending a package with a channel name.

jaimergp: This is the "path" in the anaconda.org server. e.g. concatenating this to `conda-web.anaconda.org` gives you the artifact. I'm using conda-web because of the labels; for `main` packages conda.anaconda.org would work too.

The `::` syntax is technically for MatchSpecs only, and this is more of a PackageRecord. Not that it matters much:

- jaimergp/label/something/noarch/package-1.2.3-0.tar.bz2
- jaimergp/label/something/noarch::package-1.2.3-0.tar.bz2

One more character and the opportunity for YAML to parse that as a weird dict :P I don't see much of advantage, really.

beckermr (approved):

PR #1056, opened 2024-08-27
Mark vs 19.41 as broken

isuruf:
See https://github.com/conda-forge/vc-feedstock/pull/82#discussion_r1731979252

cc @conda-forge/vc

PR #1062, opened 2024-08-30
Mark vs 19.41 as broken (again)

jdblischak:
## Checklist:

---

The transition from Visual Studio 17.10 to 17.11 for the Azure image windows-2022 has been extremely frustrating. At least twice now they have migrated to Visual Studio 17.11 but then rolled back to 17.10. These decisions are not made publicly. The only public facing PR is to update the documentation, which has been merged, even though the image has not been deployed to Azure yet (https://github.com/actions/runner-images/pull/10490#issuecomment-2321358956).

It's not possible to pin the image version (https://github.com/actions/runner-images/discussions/3464), and apparently there is no way for the conda-forge vc recipe to install the correct version based on the available version of Visual Studio (https://github.com/conda-forge/vc-feedstock/pull/82#issuecomment-2315619084). Thus our only option is to mark the latest `vs2022_win-64` as broken or not broken depending on what Azure image version is deployed on any given day (#1056, #1058).

Our win-64 builds are currently blocked by this (https://github.com/conda-forge/tiledb-feedstock/issues/340), and there is no indication that Azure will deploy the new image anytime soon. Thus I propose marking them as broken (yet again) to unblock our builds. I can send a follow-up PR to mark them as unbroken once Azure finally deploys the new image long term.

cc: @isuruf, @h-vetinari, @conda-forge/vc

### DIFF
--- a/requests/vs.yml
+++ b/requests/vs.yml
@@ -1,0 +1,6 @@
+action: broken
+packages:
+- win-64/vs2022_win-64-19.41.34120-h72b6792_20.conda
+- win-arm64/vs2022_win-64-19.41.34120-h2222f22_20.conda
+- win-64/vs2022_win-arm64-19.41.34120-hc5d368c_20.conda
+- win-arm64/vs2022_win-arm64-19.41.34120-h067b14d_20.conda


isuruf:
You can also do
```
cxx_compiler:  # [win]
  - vs2022        # [win]
cxx_compiler:  # [win]
  - 19.40          # [win]
```

for now

jdblischak:
@isuruf Thanks! I will use this workaround now to unblock my failing builds. I'll let you and others decide whether or not to merge this PR. Without knowing how soon Azure will transition, it's hard to know if it will be worth it.

jdblischak:
Just realized that those are duplicate keys for `cxx_compiler`, which results in `vs2022` being replaced by `19.40` in the `.ci_support/` YAML file. Should the second key be `cl_version` instead?

```yaml
cxx_compiler:  # [win]
  - vs2022        # [win]
cl_version:  # [win]
  - 19.40          # [win]
```

isuruf:
Fixed now. `cxx_compiler_version`

jdblischak:
Converted to draft. I applied the workaround to 2 PRs. One passed because it was still using the old image version, but the other failed (started minutes apart) because it used the new image version.

mgovers:
This relates to https://github.com/conda-forge/vc-feedstock/pull/82#discussion_r1741942869 and https://github.com/conda-forge/vc-feedstock/issues/83 . The real workaround would be to add VC build tools version `14.41.34120` to the list of supported build tools in https://github.com/conda-forge/vc-feedstock . For a sustainable fix, my guess is that https://github.com/conda-forge/vc-feedstock either needs either a major overhaul, or some additional automation.

jdblischak:
Closing now. No longer needed because the Azure windows-2022 image has finally migrated to Visual Studio 17.11 via the latest update [20240912.1.0](https://github.com/actions/runner-images/pull/10608) (it's been live for multiple days now, so I'm hoping they don't roll it back like they repeatedly did with 20240829.1.0)

PR #1074, opened 2024-09-08
feat: add allow list for feedstock autoreg

beckermr:
This PR adds an allow list for automatically registering new feedstock outputs. I have not added an admin request job because this should be rare.


beckermr:
I added a few items to start. We can add more over time as we hit things. They are hard to track down a priori.

PR #1075, opened 2024-09-09
Expand globs for known versioned ouputs

h-vetinari:
Follow-up from #1074 
> @beckermr: I added a few items to start. We can add more over time as we hit things. They are hard to track down a priori.

There's a bunch more using [this search](https://github.com/search?q=org%3Aconda-forge+%2F-+name%3A%5Cs%5Cw%2B%5C%7B%5C%7B%2F+path%3Arecipe%2Fmeta.yaml&type=code&p=1), but for now I'm not adding all these (for `vc14_runtime`, `vs2022_win-64`, etc. I didn't add it because it should be extremely rare to touch this)


beckermr (approved):
Thank you!

PR #1085, opened 2024-09-18
Delete pysimplegui 5

xhochy:
cc @conda-forge/pysimplegui 

See the reasoning here: https://github.com/conda-forge/pysimplegui-feedstock/issues/83

Keeping this simply as broken could bring us into legal trouble. If this PR is merged, I would manually delete the package as we don't have automation for it.

@conda-forge/core Deleting a package is drastic but required from my PoV. Let's get more than one approval here given its drastic nature (and to keep a positive mood in core).

### DIFF
--- a/requests/pysimplegui.yml
+++ b/requests/pysimplegui.yml
@@ -1,0 +1,3 @@
+action: broken
+packages:
+  - noarch/pysimplegui-5.0.4-pyhd8ed1ab_0.conda


ocefpaf (approved):

h-vetinari (approved):
It has ~1100 downloads over the month or so it's been online

image

So in addition to the license woes (its [restrictions](https://github.com/PySimpleGUI/PySimpleGUI/blob/08215b97a9bd5348548e000cc73d335779179e06/LICENSE.txt#L189-L209) are impossible to satisfy for us; see also discussion on [DPO](https://discuss.python.org/t/pysimplegui-now-requires-a-paid-license-opinions/48790/29) and references therein) which would already be sufficient to justify deletion IMO, it doesn't even have a ~noticeable blast radius. So LGTM.

jaimergp (approved):

PR #1091, opened 2024-09-21
Mark some builds of cppyy-cling as broken on macOS

saraedum:
cppyy-cling had insufficient testing and it turns out that these builds (3-5) are not functional at all, they just segfault or crash in other ways. This is due to a mismatch in libcxx versions used. However, no matter which libcxx version we install, these do not become functional, so this cannot be patched with a metadata patch.

The problem has since been addressed in newer builds (6 & 7.) Earlier builds (0-2) are a bit flaky but mostly work with a stricter pin on libcxx so I'll create a request for a metadata change to fix those. 

## Checklist:

  * [x] Pinged the team for the package for their input. Yes, I think that @isuruf, the other maintainer, is to some extent aware of the trouble here.


saraedum:
@conda-forge/core, this is ready for review :)

xhochy (approved):

PR #1094, opened 2024-09-30
mark noarch mpi metapackages broken

minrk:
these cause problems in conda due to a bug in conda not handling x.y versions in noarch packages

https://github.com/conda-forge/mpi-feedstock/issues/17

The `1.0.1` builds don't appear to have this problem

cc @leofang

### DIFF
--- a/requests/mpi-broken.yml
+++ b/requests/mpi-broken.yml
@@ -1,0 +1,8 @@
+action: broken
+packages:
+- noarch/mpi-1.0-msmpi.conda
+- noarch/mpi-1.0-mvapich.conda
+- noarch/mpi-1.0-openmpi.conda
+- noarch/mpi-1.0-impi.conda
+- noarch/mpi-1.0-mpi_serial.conda
+- noarch/mpi-1.0-mpich.conda


minrk:
Are broken packages supposed to retain the `main` label? The [mpi builds](https://anaconda.org/conda-forge/mpi/files) are marked `broken` but still also have the `main` label. I'm not sure when they should stop showing up in `conda install` results. I can currently install with both broken and non-broken labels (each install in a fresh `condaforge/miniforge3` image):

```
command conda create -n mpi --override-channels -c conda-forge/noarch mpi==1.0=openmpi
```

and

```
command conda create -n mpi --override-channels -c conda-forge/label/broken mpi==1.0=openmpi
```

xhochy:
They will keep the main label in the UI but vanish from the main repodata.json to ensure that they can still be installed in reproducible environments like explicit lockfiles.

beckermr:
It takes a while for the changes to propagate. We have to build the repodata patches and then wait likely two cdn cycles. That is close to an hour.

minrk:
Thanks, I'm never sure when to expect repodata to be up to date.

PR #1106, opened 2024-10-18
Unarchive XZ feedstock

timostrunk:
The xz feedstock was archived quite some time ago during the code-injection security issue. This has been resolved in version 5.6.2. Since then 5.6.3 was released, which contains another windows cli security fix:  https://github.com/tukaani-project/xz/releases. I think we should start building releases here again (and also maybe change the sourcecode source in the feedstock).

I am not part of the xz-feedstock team and no expert on xz.

* [ ] I want to unarchive a feedstock:
  
 ping @conda-forge/xz

### DIFF
--- a/requests/unarchive_xz.yml
+++ b/requests/unarchive_xz.yml
@@ -1,0 +1,3 @@
+action: unarchive
+feedstocks:
+  - xz


beckermr:
cc @conda-forge/core This may need some discussion?

xhochy:
I would like to take this an opportunity to split the `xz` package into library and utils (which would be quite helpful from a licensing perspective). I'm happy to step in as a maintainer.

beckermr:
Ok @xhochy. I'll defer merging this pr to you then so you can proceed at your own pace.

h-vetinari (approved):
Great to hear @xhochy! In any case, the feedstock will need to be unarchived to raise PRs again

PR #1123, opened 2024-10-29
chore: reset token for repodata patches

beckermr:
This PR resets the tokens for the repodata patches so we can run it on gha.


requests/cfrp.yml
```diff
@@ -0,0 +1,3 @@
+action: token_reset
+feedstocks:
+  - conda-forge-repodata-patches-feedstock
```
beckermr, line 3:
```suggestion
  - conda-forge-repodata-patches
```

PR #1152, opened 2024-11-08
Add intel-pti package

ZzEeKkAa:
## Checklist:

ping @conda-forge/pti-gpu-unitrace 

Intel recently release `intel-pti` package in https://software.repos.intel.com/python/conda channel. This package is important cause it is required dependency for the intel's gpu (aka xpu) backend for pytorch. Instead of adding it as a repack at https://github.com/conda-forge/intel_repack-feedstock it is possible to build it from source here:
https://github.com/conda-forge/pti-gpu-unitrace-feedstock/pull/1

`pti-gpu-unitrace` does not have independent release cycle, so we will get regular updates on this as well.


ZzEeKkAa:
Thank you!

PR #1168, opened 2024-11-17
enable clang_impl outputs on ctng-compiler-activation

h-vetinari:
Now that we have CFEP-24, the new `clang{,xx}_impl` outputs from https://github.com/conda-forge/ctng-compiler-activation-feedstock/pull/131 are failing validation. Add them.

PR #1169, opened 2024-11-17
clang_impl, take 2

h-vetinari:
Redo #1168

### DIFF
--- a/requests/ctng.yml
+++ b/requests/ctng.yml
@@ -1,0 +1,8 @@
+action: add_feedstock_output
+feedstock_to_output_mapping:
+  - ctng-compiler-activation: clang_impl_linux-64
+  - ctng-compiler-activation: clang_impl_linux-aarch64
+  - ctng-compiler-activation: clang_impl_linux-ppc64le
+  - ctng-compiler-activation: clangxx_impl_linux-64
+  - ctng-compiler-activation: clangxx_impl_linux-aarch64
+  - ctng-compiler-activation: clangxx_impl_linux-ppc64le


requests/ctng.yml
```diff
@@ -0,0 +1,8 @@
+action: add_feedstock_output
+feedstock_to_output_mapping:
+  - ctng-compiler-activation: clang_impl_linux-64
+  - ctng-compiler-activation: clang_impl_linux-aarch64
+  - ctng-compiler-activation: clang_impl_linux-ppc64le
+  - ctng-compiler-activation: clangxx_impl_linux-64
+  - ctng-compiler-activation: clangxx_impl_linux-aarch64
+  - ctng-compiler-activation: clangxx_impl_linux-ppc64le
```
h-vetinari, line 8: @beckermr, I cannot from looking at the example if there's a way for adding several outputs to a given feedstock allowlist?
The list approach in #1168 failed, and the approach here also doesn't work, because (I realized too late) the key-value pairs get overridden - so https://github.com/conda-forge/feedstock-outputs/commit/446a061f775844e56b3073136f9e67e2368c39d3 only added `clangxx_impl_linux-ppc64le`.

I think we should add:
* a canonical & documented way for adding multiple outputs to a feedstock
* some schema validation so that mistakes like I did in #1168 and this PR get caught in CI here.

Thoughts?

h-vetinari: OK, actually, this PR did end up working, I just checked too early and only saw one commit. In the end one commit per line here was [created](https://github.com/conda-forge/feedstock-outputs/commits/446a061f775844e56b3073136f9e67e2368c39d3/)

beckermr: Yep. That's why we use a list and not a dict.

h-vetinari: > Yep. That's why we use a list and not a dict.

It seems to be using a list-of-dict where dict is feedstock-to-output. If that's the target picture, then we should fail schema validation if someone does (i.e. #1168) list-of-dict-from-feedstock-to-LIST-of-outputs.

Or alternatively, allow that case (ensuring that it works, of course), e.g.
```yaml
action: add_feedstock_output
feedstocks:
  - feedstock-name1: output1
  - feedstock-name2:
    - output2a
    - output2b
```
which I think would be more intuitive.

Relatedly, I'd also prefer if adding a glob was under a different action (e.g. `add_feedstock_output_rule`), because right now the example doesn't indicate what actually makes the difference between a single output and a glob rule (probably the presence of `*`?).

beckermr: Feel free to make prs for features you'd like to see.  We use the presence of any glob characters to detect globs.

PR #1170, opened 2024-11-18
Request package output change for `flask-security`

nicoddemus:
https://github.com/conda-forge/flask-security-feedstock/pull/7

Perhaps the upload problem is due that the package has changed to `noarch`?

https://github.com/conda-forge/admin-requests/pull/1167 attempted to reset the token, but that did not work.

### DIFF
--- a/requests/flask-security.yml
+++ b/requests/flask-security.yml
@@ -1,0 +1,3 @@
+action: add_feedstock_output
+feedstock_to_output_mapping:
+  - flask-security: "flask-security*"
--- a/requests/flask-security.yml
+++ b/requests/flask-security.yml
@@ -1,3 +1,3 @@
 action: add_feedstock_output
 feedstock_to_output_mapping:
-  - flask-security: "flask-security*"
+  - flask-security: "flask-security"


requests/flask-security.yml
```diff
@@ -0,0 +1,3 @@
+action: add_feedstock_output
+feedstock_to_output_mapping:
+  - flask-security: "flask-security*"
```
nicoddemus, line 3: The failure output is:

```
validation results:
{
  "noarch/flask-security-5.5.2-pyhff2d567_0.conda": false
}
```

beckermr (changes requested):
This PR is not needed. We only check the name. Someone from core will need to debug likely.

requests/flask-security.yml
beckermr, line 3:
```suggestion
  - flask-security: "flask-security"
```

beckermr (approved):

beckermr:
I checked the feedstock outputs repo and this package is missing. It must have not been registered at some point, but I don't know why. In any case, merging this pr should fix it.

nicoddemus:
Thanks!

PR #1177, opened 2024-11-20
Add request for ctapipe-base

maxnoe:
## Checklist:

* [X] I want to add a package output to a feedstock:
  * [X] Pinged the relevant feedstock team(s)
  * [X] Added a small description of why the output is being added.

We want to split our package `ctapipe` into a `ctapipe-base` package only depending on the core required dependencies and a `ctapipe` that comes with additional optional dependencies.

See PR here: https://github.com/conda-forge/ctapipe-feedstock/pull/31/files


maxnoe:
Thanks a lot for the very quick resolution!

PR #1193, opened 2024-11-26
prepare to split suitesparse outputs

minrk:
each library will get its own output, including the addition of suitesparse-mongoose not previously included:

https://github.com/conda-forge/suitesparse-feedstock/pull/98

package names match those in ubuntu (not including the ABI suffix, so libamd, not libamd3)


h-vetinari:
Been wanting to do this (and switch to CMake, which gained first-class upstream support) for a long time - thanks  a lot!

PR #1199, opened 2024-11-30
add rdkit cleanup yml

greglandrum:
I would like to remove the artifacts from two old releases of the RDKit project from the archive.

In the early days of the RDKit feedstock we used a different version naming scheme. Unfortunately whatever logic is being used to pick the most recent version to show considers that newer than the actual versions. It would be nice if that landing page showed the actual current version.

image

This makes it look like the project hasn't been updated since 2017 and is a poor user experience.

The two releases that I suggest marking as broken are seven years old.

I am one of the maintainers of the RDKit feedstock and the main developer of the project itself.

## Checklist:

### DIFF
--- a/requests/rdkit-cleanup.yml
+++ b/requests/rdkit-cleanup.yml
@@ -1,0 +1,21 @@
+action: broken
+packages:
+- osx-64/rdkit-Release_2017_09_3-py36_3.tar.bz2
+- osx-64/rdkit-Release_2017_09_3-py35_3.tar.bz2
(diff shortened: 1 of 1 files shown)


greglandrum:
Hi @conda-forge/core , could you please give a thumbs up or down on this one?

xhochy (approved):
This might not change the landing page, but OK to get rid of them because of the `Release_` prefix in the version number.

greglandrum:
Thanks @xhochy!

PR #1204, opened 2024-12-02
mark rasterio 1.4.0 build=0 as broken

akrherz:
## Checklist:

The build=0 of `rasterio=1.4.0` (back in September) with conda-forge/rasterio-feedstock#316 ended up with an invalid `gdal` dependency of just `gdal` without any version pinning, [for example](https://conda-metadata-app.streamlit.app/?q=conda-forge%2Flinux-64%2Frasterio-1.4.0-py312heb92005_0.conda)

We got this straightened out a day later with build=1 and conda-forge/rasterio-feedstock#318

Sadly, this build=0 is causing some grief now conda-forge/rasterio-feedstock#325 with the mamba solver using this build to satiate a `gdal=3.10` dependency.

Since build=1 immediately fixed this, I think it is best to just mark build=0 as broken.

ping @conda-forge/rasterio


hmaarrfk:
Are you sure you want this instead of repo patches?

akrherz:
> Are you sure you want this instead of repo patches?

Fair question.  The build logs expired, so I can't fully inspect what happened with that build and build=1 came immediately after it, so it should be effectively the same thing, but it is difficult for me to verify.

hmaarrfk:
ok do consider it next time.

PR #1019, opened 2024-06-17
mark minzip 4.0.7 on windows as broken

akrherz:
## Checklist:

* [x] I want to mark a package as broken (or not broken):
  * [x] Added a description of the problem with the package in the PR description.
  * [x] Pinged the team for the package for their input.

The windows build of minizip 4.0.7 changed a .dll file name conda-forge/minizip-feedstock#17 and is causing downstream build grief on conda-forge.  @ocefpaf asked that I mark it as broken for now

PR #1219, opened 2024-12-12
mark minizip windows 4.0.7-h9fa1bad_1 as broken

akrherz:

What is old is new again, same as #1019, the updated minizip build for windows is broken as it has a different name for the dll, which breaks things left and right :(

cc: @rouault

ping @conda-forge/minizip


ocefpaf:
Somehow the hash of that package predicted this: "4.0.7-h9fa1**bad** _1"

PR #1249, opened 2024-12-25
reset feedstock token for behave

stuertz:
Request token reset for behave-feedstock:

https://github.com/conda-forge/behave-feedstock/issues/11

### DIFF
--- a/requests/behave.yml
+++ b/requests/behave.yml
@@ -1,0 +1,7 @@
+action: token_reset
+feedstocks:
+- behave
+# can be a list with azure, travis, circle, drone, appveyor, github_actions
+skip_providers: []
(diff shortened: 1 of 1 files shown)


requests/behave.yml
```diff
@@ -0,0 +1,7 @@
+action: token_reset
+feedstocks:
+- behave
+# can be a list with azure, travis, circle, drone, appveyor, github_actions
+skip_providers: []
+# we can expire tokens earlier if desired - default is null which is immediately
+# existing_tokens_time_to_expiration: null  # set to an int in seconds to use
```
ocefpaf, line 7: Are these lines needed for a token reset for all? My guess is that you only need it if you are doing  a specific service, no?

PR #1329, opened 2025-01-21
Add `parsl-with-*` extras

xylar:
We have `parsl-with-*` packages for many `parsl` extras already.  A new one was added in a recent update, `parsl-with-globus_compute`:
https://github.com/conda-forge/parsl-feedstock/pull/115

In general, we want to keep up with new extras as they are added to parsl.

## Checklist:

ping @conda-forge/parsl

### DIFF
--- a/requests/add-parsl-output.yml
+++ b/requests/add-parsl-output.yml
@@ -1,0 +1,3 @@
+action: add_feedstock_output
+feedstock_to_output_mapping:
+  - parsl: "parsl-with-*"


xylar:
@conda-forge/core, I believe this is ready for your review when you have time.

requests/add-parsl-output.yml
```diff
@@ -0,0 +1,3 @@
+action: add_feedstock_output
+feedstock_to_output_mapping:
+  - parsl: "parsl-with-*"
```
ocefpaf, line 3: I don't know if the `*` glob works here. Let's ask @beckermr, who create these yaml requests, to be sure before merging.

xylar: @ocefpaf, I don't mind getting confirmation but I have done this in the recent past and it does work. It is also in the example yaml this way.

xylar: More specifically, here is the example:
https://github.com/conda-forge/admin-requests/blob/main/examples/example-add-feedstock-output.yml#L6
And here hare similar changes I have made for 2 other feedstocks:
https://github.com/xylar/admin-requests/blob/f84779f6de0c4dd5f0de582d56dc4a80ac8384f8/requests/wetterdienst.yml
https://github.com/xylar/admin-requests/blob/381528a2fb6850d68ea4f94f9fc3a673b0d6b829/requests/databricks-sql-connector-add-extras.yml

ocefpaf: Your expertise works as confirmation! I'm way behind the new stuff in conda-forge and clearly you know what you are doing. Thank for teaching me that one!

xylar: No problem!

PR #1344, opened 2025-01-25
allow multiple outputs in ncrystal-feedstock

tkittel:
In order to split the C++/library parts of NCrystal from the Python layer, the ncrystal-feedstock needs to be modified to use multiple outputs. So before there was just an "ncrystal" package, but in the future there will be at least "ncrystal-core", "ncrystal-python",  and "ncrystal" packages, where the last will be a metapackage depending on the rest and ensuring synchronisation of versions. In the future the ncrystal-core itself might be split further (into libs, headers, data files), but that is not going to happen in the short term.

I am the maintainer of the ncrystal-feedstock, so I count myself as having been pinged :-)

### DIFF
--- a/requests/add_ncrystal_outputs.yml
+++ b/requests/add_ncrystal_outputs.yml
@@ -1,0 +1,3 @@
+action: add_feedstock_output
+feedstock_to_output_mapping:
+  - ncrystal: "ncrystal-*"


h-vetinari:
Why not call the C++ library `libncrystal`?

tkittel:
@h-vetinari mostly because it is not 100% just the lib, but more importantly to follow 1-to-1 the naming of the corresponding pypi packages.

PR #1374, opened 2025-02-10
Add static outputs to collier and ninja-hep-ph

matthewfeickert:
* To work properly as dependencies of mg5amcnlo (https://github.com/conda-forge/mg5amcnlo-feedstock) add static library outputs to the collier (https://github.com/conda-forge/collier-feedstock) and ninja-hep-ph (https://github.com/conda-forge/ninja-hep-ph-feedstock) feedstocks.

## Guidelines for marking packages as broken:

* We prefer to patch the repo data (see [here](https://github.com/conda-forge/conda-forge-repodata-patches-feedstock))
  instead of marking packages as broken. This alternative workflow makes environments more reproducible.
* Packages with requirements/metadata that are too strict but otherwise work are
  not technically broken and should not be marked as such.
* Packages with missing metadata can be marked as broken on a temporary basis
  but should be patched in the repo data and be marked unbroken later.
* In some cases where the number of users of a package is small or it is used by
  the maintainers only, we can allow packages to be marked broken more liberally.
* We (`conda-forge/core`) try to make a decision on these requests within 24 hours.

What will happen when a package is marked broken?

* Our bots will add the `broken` label to the package. The `main` label will remain on the package and this is normal.
* Our bots will rebuild our repodata patches to remove this package from the repodata.
* In a few hours after the `anaconda.org` CDN picks up the new patches, you will no longer be able to install the package from the `main` channel.

## Checklist:

* [ ] I want to mark a package as broken (or not broken):
  * [ ] Added a description of the problem with the package in the PR description.
  * [ ] Pinged the team for the package for their input.

* [ ] I want to archive a feedstock:
  * [ ] Pinged the team for that feedstock for their input.
  * [ ] Make sure you have opened an issue on the feedstock explaining why it was archived.
  * [ ] Linked that issue in this PR description.
  * [ ] Added links to any other relevant issues/PRs in the PR description.

* [ ] I want to request (or revoke) access to an opt-in CI resource:
  * [ ] Pinged the relevant feedstock team(s)
  * [ ] Added a small description explaining why access is needed

* [ ] I want to copy an artifact following [CFEP-3](https://github.com/conda-forge/cfep/blob/main/cfep-03.md):
  * [ ] Pinged the relevant feedstock team(s)
  * [ ] Added a reference to the original PR
  * [ ] Posted a link to the conda artifacts
  * [ ] Posted a link to the build logs

* [x] I want to add a package output to a feedstock:
  * [x] Pinged the relevant feedstock team(s)
[...]

PR #1376, opened 2025-02-11
Set pull.rebase = false for global git configs

jaimergp:
This is supposed to address the issues observed in https://github.com/conda-forge/admin-requests/pull/1374. We had this line in other workflows, so let's add it everywhere for consistency.

### DIFF
--- a/.github/workflows/create_feedstocks.yml
+++ b/.github/workflows/create_feedstocks.yml
@@ -30,6 +30,7 @@ jobs:
         run: |
           git config --global user.email "chrisburr73+conda-forge-admin@gmail.com"
           git config --global user.name "conda-forge-admin"
+          git config --global pull.rebase false
           git add *
           git commit -am "make sure we have no windows line endings" || exit 0
           for i in `seq 1 5`; do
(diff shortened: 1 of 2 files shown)


beckermr (changes requested):
I do not understand what this PR is supposed to fix. The issue we had was that thw request was misformatted and I missed that in my review. @chrisburr fixed it in this [commit](https://github.com/conda-forge/admin-requests/commit/bbf668655024ddccb994cf0f1b0454593e912c97).

beckermr:
The repo retries requests over and over again by default, which is why the job was on a loop.

jaimergp:
I see. I misunderstood then. Yet, the error observed in this log https://github.com/conda-forge/admin-requests/actions/runs/13252868280/job/36994275445 could be fixed by this PR I think? Might be orthogonal to the issue I linked to, sorry for the confusion.

beckermr:
No need to be sorry at all! Did that error happen more than once? I think it might be a race condition that resolves itself on the next attempt?

jaimergp:
Probably. Still, git could have maybe worked around that conflict with a different merge strategy instead of fast-forward. Not sure if we want to allow that either. Feel free to close.

PR #1390, opened 2025-02-14
Add sdl2 and sdl3 outputs in sdl feedstock

ymontmarin:
This PR adds the `sdl2` and `sdl3`output to the [sdl-feedstock](https://github.com/conda-forge/sdl-feedstock).

To follow package manager habits, SDL major version are released with different package name, but following discussion in https://github.com/conda-forge/staged-recipes/pull/29061#issuecomment-2658421352 we manage to reunite all the major version in one feedstock.

## Checklist:

### DIFF
--- a/requests/sdl.yml
+++ b/requests/sdl.yml
@@ -1,0 +1,4 @@
+action: add_feedstock_output
+feedstock_to_output_mapping:
+  - sdl: sdl2
+  - sdl: sdl3


requests/sdl.yml
```diff
@@ -0,0 +1,4 @@
+action: add_feedstock_output
+feedstock_to_output_mapping:
+  - sdl: sdl2
+  - sdl: sdl3
```
h-vetinari, line 4: Do we need `sdl2-compat` here as well?

ymontmarin: No we do not as `sdl2-compat` upstream package is intended to be a new `sdl2` package with patch number >=50 as discussed here: https://github.com/conda-forge/staged-recipes/pull/29061#issuecomment-2653903072

PR #1396, opened 2025-02-19
Remove BasicTools feedstock. Change name of lib to Muscat

fbordeu:

## Checklist:

* [x] I want to archive a feedstock:
  * [x] Pinged the team for that feedstock for their input.
  * [x] Make sure you have opened an issue on the feedstock explaining why it was archived.
  * [x] Linked that issue in this PR description.
  * [x] Added links to any other relevant issues/PRs in the PR description.

### DIFF
--- a/requests/basictools-archive.yml
+++ b/requests/basictools-archive.yml
@@ -1,0 +1,3 @@
+action: archive
+feedstocks:
+  - basictools


fbordeu:
Pinged the team for that feedstock for their input. -> I'm the maintainer.

fbordeu:
Make sure you have opened an issue on the feedstock explaining why it was archived.

https://github.com/conda-forge/basictools-feedstock/issues/55

jaimergp:
By archiving it, the 1.x series won't be rebuilt for newer libraries like boost, so it might become increasingly difficult to install in the future. Just confirming you are ok with that.

fbordeu:
Yes I confirm

PR #1397, opened 2025-02-19
muscat-split-feedstock. Add GPU ci

fbordeu:
The new functionalities of muscat uses kokkos as backend. This backend can use cpu or gpu for the computation.
To test muscat packages (muscat-core package) I need access to a gpu ci. The compilation as well as the tests are really small. 

Compilation of around 60 files. les dans 10 minutes.
Test of the lib. less than 5 minutes.

## Checklist:


fbordeu:
I'm the maintainer of https://github.com/conda-forge/muscat-split-feedstock

jaimergp:
You'll need to add yourself to https://github.com/Quansight/open-gpu-server/blob/main/access/conda-forge-users.json as well. Ping me once done.

fbordeu:
Ok, thanks for the info. Now I'm on vacation. Will do it next week.

fbordeu:
hi @jaimergp I just created a MR to add myself to https://github.com/Quansight/open-gpu-server/blob/main/access/conda-forge-users.json  here is the MR https://github.com/Quansight/open-gpu-server/pull/59

jaimergp (approved):

fbordeu:
@jaimergp Thanks

PR #1399, opened 2025-02-20
Request additional outputs for pytorch_geometric

jamesmyatt:
To support https://github.com/conda-forge/pytorch_geometric-feedstock/pull/51. Justification and agreement from maintainers are in that PR.

## Checklist:

### DIFF
--- a/requests/pyg.yml
+++ b/requests/pyg.yml
@@ -1,0 +1,7 @@
+# To support https://github.com/conda-forge/pytorch_geometric-feedstock/pull/51
+action: add_feedstock_output
+feedstock_to_output_mapping:
+  - pytorch_geometric: pyg
+  - pytorch_geometric: pyg-modelhub
+  - pytorch_geometric: pyg-graphgym
+  - pytorch_geometric: pyg-full
--- a/requests/pytorch_geometric.yml
+++ b/requests/pytorch_geometric.yml
@@ -1,7 +1,5 @@
 # To support https://github.com/conda-forge/pytorch_geometric-feedstock/pull/51
 action: add_feedstock_output
 feedstock_to_output_mapping:
-  - pytorch_geometric: pyg
-  - pytorch_geometric: pyg-modelhub
-  - pytorch_geometric: pyg-graphgym
-  - pytorch_geometric: pyg-full
+  - pytorch_geometric: pytorch_geometric-modelhub
+  - pytorch_geometric: pytorch_geometric-graphgym


hmaarrfk:
I'm not sure we want to strive for package name parity with all channels.

The https://pypi.org/project/pyg/ project exists on pypi and we typically like to stay closer to pypi if we can.

So this seems like 👎🏾 from me.

beckermr:
The pypi project has been declared dead. Here is a direct quote. 

"==========================

This project is now **DEAD**.
It was born with the intent to provide a better (IMHO) alternative to Pip.
Since now Pip has got a lot of great improvements, this is not needed anymore.

So the advice here is to use Pip instead. Pyg has not been maintained for over
2 years now. It may well be broken in many places. The code is old and would
need a refactoring, but there is not need anymore."

beckermr:
So I'm a definite plus one.

beckermr (approved):

h-vetinari:
> * Adds pyg-graphgym package with parity for "torch_geometric[graphgym]" pypi package
> * Adds pyg-modelhub package with parity for "torch_geometric[modelhub]" pypi package:
> * Adds pyg-full package with parity for "torch_geometric[full]" pypi package

Why are these not named `torch_geometric-graphgym`, etc.? IMO there's good reason to match the PyPI naming (to the degree that we can and deal with extras in general). I don't see why we should use a completely different name just because another conda channel uses it.

> Adds pyg subpackage with parity for:
> *  "torch_geometric" pypi package: [...]
> *  "pyg" and "pytorch-geometric" packages in "pyg" channel: [...]

Why three different names for the same thing?! This makes no sense to me. FWIW, since I see it's from the same upstream Github organization, we already have https://github.com/conda-forge/pyg-lib-feedstock

beckermr:
I understand your points @h-vetinari. 


Historically the core team has left the namespace as a first-come-first-serve resource. 

We've only, to my knowledge, interceded in this process in a handful of times in the following ways. 

1. Specific package names around low-level tools (ie build versus python-build, timezone data, etc)

2. The handling of static libraries since their presence can have adverse effects. 

I could be convinced to agree that we don't want a proliferation of arbitrary package aliases lying around. However, I'd like to see that rule passed as a CFEP. We currently have quite a few dash versus underscore situations, our compilers carry similar aliases, and so I think we need to understand how this kind of rule might work. 

Finally, the activity of pyg-lib or even the presence of pyg on pypi really doesn't have a bearing on this decision in the first place. 

The authors of those packages are not here right now requesting the names.

beckermr:
Again I don't disagree with any of your points strongly, but I don't know how to make a call between various possible aliases in terms of which ones are OK and which ones are not. 

> The authors of those packages are not here right now requesting the names.

All I meant here is that I don't see a reason to not honor the "first come, first serve" rule we've always had.

mikemhenry:
I will just add that I would be happy with a subset of these output if that helps @h-vetinari since I do agree that we don't need to have every possible combination but I think it will help with users find these packages

hmaarrfk:
> All I meant here is that I don't see a reason to not honor the "first come, first serve" rule we've always had.

I guess you are right.

hmaarrfk (approved):

h-vetinari:
> I will just add that I would be happy with a subset of these output if that helps @h-vetinari since I do agree that we don't need to have every possible combination but I think it will help with users find these packages

You have convinced two core members, I won't stand in the way here.

beckermr:
Thank you @h-vetinari. :) 

@mikemhenry what is your proposal for which aliases to keep and which ones we would not keep?

mikemhenry:
I've signed off for today but will comment tomorrow 😀

jamesmyatt:
Thanks everyone for reviewing this. For context (summarised from below, which is TL;DR):

* "pyg" is what the upstream project calls itself: https://pyg.org/. "pyg-lib" is a recent (sub)project that is part of the same upstream project.
* "pyg" is the official conda channel belonging to the upstream project, and provides the "pyg" package: https://pytorch-geometric.readthedocs.io/en/latest/install/installation.html
* The conda-forge "pytorch_geometric" package matches the "torch_geometric[full]" pypi package, except that one of the dependencies isn't on conda-forge yet and one can only be installed on linux.
* **The primary objective** is to provide a conda-forge package that matches the core "torch_geometric" pypi package and (by extension) the official "pyg::pyg" package.
* I'm not convinced that installing the "full" extras is a good idea for the vast majority of situations: it looks to me like a convenience for someone who is exploring graph learning.
* "pyg-graphgym" is just shorter than "pytorch_geometric-graphgym" and unlikely to cause any confusion.

So I'm happy with any or all of the following solutions:

* Accept my original proposal.
* Change the "pytorch_geometric" output to match the basic pypi package without extras, which is a _breaking change_, but I suspect that not many people will notice, since they're probably using the pyg channel instead.
* Change the sub-packages to "pytorch_geometric-core", "pytorch_geometric-graphgym", etc. and don't provide any "pyg-*" packages.
* Drop the "full" subpackage and provide only the "core", "graphgym" and "modelhub" ones.

-- 

More details on the context:

* Upstream:
  * The website for the upstream project is https://pyg.org/ so "pyg" is a well established name associated with the project. It talks about itself using both "PyG" and "PyTorch Geometric" interchangably.
  * "pyg" is the official conda channel associated with the upstream project and they publish the package as "pyg" there, and mention it in the official installation instructions: https://pytorch-geometric.readthedocs.io/en/latest/install/installation.html
  * The upstream project also publishes an equivalent package as "pytorch-geometric" to the "pyg" channel.
  * The upstream pypi project is "torch_geometric" but there's a long-standing conda convention of the pytorch packages being called "pytorch*" when the pypi package is just "torch*", like the core pytorch library.
  * The upstream pypi package defines a few extras "graphgym", "modelhub" and "full". I can see people using the first two, but I think the last one is just a convenience for installing lots of packages relevant to graph machine learning.
  * The "pyg-lib" library is part of the same overall upstream project, but is a separate package.
* conda-forge:
  * The existing conda-forge package for this project is called "pytorch_geometric" with an underscore, rather than a hyphen. I don't think it's worth changing or providing to the hyphenated version, even if the upstream "pyg" channel does.
  * The existing "pytorch_geometric" conda-forge package is equivalent to the pypi package "torch_geometric[full]", so includes lots of optional extras. The PR maintains this mapping for consistency, even if it's rarely what people actually want to install (c.f. matplotlib and seaborn).
  * **My main objective** is to publish a conda-forge package that matches the core pypi package "torch_geometric" without extras and the "official" conda package "pyg::pyg". This has become more important now that the official "pytorch" channel is deprecated and users are pointed to the conda-forge packages instead.
  * Given that "pyg" is an established name for the project, then the sub-packages "pyg-modelhub", etc are more compact than "pytorch_geometric-modelhub", etc. But it's not really important.
  * If there are going to be some packages called "pyg-..." and some called "pytorch_geometric...", then one set should just be synonyms for the other, i.e. in this case, "pytorch_geometric" is a synonym for "pyg-full".

jamesmyatt:
> Why three different names for the same thing?!

@h-vetinari, they're 3 different upstream names for the same thing. The conda-forge name "pytorch_geometric" is actually a 4th.  But "pyg" is the current branding for the project, and it's common for the conda packages to be called "pytorch..." when the pypi package is "torch...", like Pytorch itself.

beckermr:
Thanks everyone. I think a consensus solution is to do a combination of 2+4 and to NOT add the `pyg-*`. This keeps everyone close to equally unhappy, which is a good sign.

jamesmyatt:
I've opened https://github.com/conda-forge/pytorch_geometric-feedstock/issues/52 to specifically capture the breaking change for the core package.

jan-janssen:
@jamesmyatt I like the changes, please go ahead.

beckermr (approved):
LGTM and thanks for working with us!

PR #1401, opened 2025-02-24
Add `apache-beam-with-*` extras

xylar:
The latest update to apache-beam in https://github.com/conda-forge/apache-beam-feedstock/pull/91 attempts to add outputs for new extras, but these are not being allowed at the moment.

## Checklist:

ping @conda-forge/apache-beam


xylar:
Thank you @ocefpaf!!

PR #53, opened 2020-05-16
Mark re2=2020.0501 as broken

xhochy:
They bumped the SONAME but we don't have a `run_exports` in this package at all, feedstock issue: https://github.com/conda-forge/re2-feedstock/issues/37

I'll fix the feedstock and make a repodata patch afterwards.

Checklist:

* [x] Added a link to the relevant issue in the PR description.
* [x] Pinged the team for the package.

@conda-forge/re2

PR #51, opened 2020-05-14
Mark python-language-server 0.32.0 build0 as broken

ccordoba12:
* Dependencies were not updated correctly in that build. See https://github.com/conda-forge/python-language-server-feedstock/pull/52 and https://github.com/conda-forge/python-language-server-feedstock/pull/53 for the fix.
* Ping @conda-forge/python-language-server about it (I'm part of that team).

PR #48, opened 2020-05-08
Remove broken Sphinxdft version

jan-janssen:
Checklist:

* [x] Added a link to the relevant issue in the PR description.
* [x] Pinged the team for the package.
<!
  For example if you are trying to mark a python package as broken.

      ping @conda-forge/python
!>

PR #1406, opened 2025-02-26
Mark vendoring jupyter_scheduler packages as broken

zklaus:

## Checklist:

## Description
Older builder of jupyter_scheduler vendor a significant number of other, popular packages such as SQLAlchemy and Pydantic, among others.
This happened by accident because the standard, grayskull-provided install script line made pip install dependencies, which slipped through quality control.

The feedstock side is discussed in conda-forge/jupyter_scheduler-feedstock#46, the recipe is fixed to prevent this happening again in the future (though the rebuilds of older versions are still outstanding.

Here, we need to mark the offending builds as broken, which this PR is for.

@conda-forge/jupyter_scheduler, please weigh in as you see fit.


h-vetinari:
> though the rebuilds of older versions are still outstanding.

IMO the rebuilds (to whatever extent that is deemed necessary) should come before marking the existing packages as broken.

zklaus:
> > though the rebuilds of older versions are still outstanding.
> 
> IMO the rebuilds (to whatever extent that is deemed necessary) should come before marking the existing packages as broken.

Makes sense. Let's see what the team has to say. My gut feeling is that the three remaining versions is enough.

andrii-i:
Hi everone. As discussed here https://github.com/conda-forge/jupyter_scheduler-feedstock/issues/46#issuecomment-2716140324, I've rebuilt latest patch of each minor version >2, <2.8.0. Non-rebuilt 2.x versions < 2.8.0 and all 1.x versions can be marked as broken (1.x supports JupyterLab 3 which reached end of life and end of support).

Versions I've rebuilt:
* 2.7.1 #49 
* 2.6.0 #50 
* 2.5.2 #51 
* 2.4.0 #52 
* 2.3.0 #53 
* 2.2.0 #54 
* 2.1.0 #48 

Versions OK to be marked as broken:
* 2.7.0
* 2.5.1
* 2.5.0
* All 1.x versions

Thank you for looking into this.

andrii-i (changes requested):
Only mark following versions as broken:
 - 2.7.0
 - 2.5.1
 - 2.5.0
 - All 1.x versions

PR #1412, opened 2025-03-01
add ipython output _ipython_tests

bollwyvl:

## Checklist:

ping @conda-forge/ipython

This adds an output `_ipython_tests` to `ipython`, added in https://github.com/conda-forge/ipython-feedstock/pull/231

### DIFF
--- a/requests/ipython-tests.yml
+++ b/requests/ipython-tests.yml
@@ -1,0 +1,3 @@
+action: add_feedstock_output
+feedstock_to_output_mapping:
+  - ipython: _ipython_tests


bollwyvl:
@conda-forge/core ready for review

bollwyvl:
> needs to be an output?

Nope, guess it doesn't have to. If the answer is _no_ (and that's fine) i can put it back in, but... 

> seems wasteful

Right: wasteful for whom? It's pretty much same amount of storage, maybe a little worse from a compression point of view. Certainly this will make another 2 packages per release, so the rate of repodata growth will increase. However, this cuts 30% off the download/install size for `ipython` users. 

After spending a couple years trying to get it out of the pypi wheel (mostly for in-browser consumption), was trying to realize some gains here as well, and had seen some other suites that created unvendored test packages.

bollwyvl:
I guess some other ways forward:
- download the tests at test time (ick)
- lobby for a more configurable, per-output `channel_targets`, such that it could be uploaded to e.g. `label/ipython_tests`

bollwyvl:
Thanks again for the review: hoisted this up to zulip, and will roll back the split on the feedstock PR, maybe we can get some more discussion.

beckermr:
I've been meaning to add a feature that allows folks to skip uploads for some packages via a config option in the conda-forge.yml. 

That seems like it'd be a straight forward solution here. 

Downloading the tests at test some seems fine to me fwiw. Idk why that option isn't ok either?

beckermr:
My long-term concerns are about storage costs / size for these kinds of uses for the index packages. Having the tests on the internet is no worse than having the package source there, which we also do not store.

bollwyvl:
Reframed some thoughts here: https://github.com/conda-forge/conda-smithy/issues/2263

PR #1433, opened 2025-03-12
Mark versions 11.1.0 to 11.1.4 of fmt as broken due to the ABI breakage

SylvainCorlay:
As reported by @ryanvolz in https://github.com/conda-forge/fmt-feedstock/issues/61, fmt 11.1 broke is not ABI compatible with version 11.0.

This has had an impact on several downstream projects, such as `spdlog`, `mamba`, and `arcticdb`.

In this PR, I mark all corresponding builds of fmt as broken.

cc @h-vetinari @jjerphan.


jjerphan (commented):
Thank you, Sylvain. 

Although this LGTM, I am thinking of the implications for packages which have been built against those builds of fmt>=11.1 since they might have been used for weeks or months in practice. 

I think we might need to perform a few operations before marking those builds as broken, including:
 - changing runtime requirements of fmt from major to minor versions
 - performing a migration for a rebuild against fmt 11.0
 - pinning fmt 11.0 to conda-forge global pinning

I let @h-vetinari (who has more authority) confirm or correct this remark.

h-vetinari:
The problem is that https://github.com/conda-forge/fmt-feedstock/pull/60 has been online for a week already, and there are a bunch of packages already depending on it
```
$ mamba repoquery whoneeds -c conda-forge fmt | grep "fmt >=11.1" | wc -l
53
```
(and potentially more that depend on those new builds/versions transitively).

With the exception of tiledb (is that broken too?), most of these have very few downloads though, so it may be the least bad choice here. 🤷

SylvainCorlay:
So, if we do this, it's got to be as soon as possible I guess...

beckermr:
If we want to permanently pin down to the minor version, then I think we need to do following.

- mark fmt 11.1 broken (this PR)
- go through existing packages and adjust any fmt pins to be pinned down to the minor version
- push new versions of fmt 11.0 and 11.1 with the correct run exports
- update the global pinnings to pin down to the minor version
- mark older builds of fmt 11.0 with the wrong run_exports as broken

cc @conda-forge/core for viz and advice

SylvainCorlay:
> push new versions of fmt 11.0 and 11.1 with the correct run exports

So, operate as if fmt's contract was to only be ABI compatible on patch releases.

SylvainCorlay:
On a side note, version `11.1.4` was released upstream to fix an ABI issue that impacted `11.1.0` to `11.1.3` on windows (https://github.com/fmtlib/fmt/issues/4359). This seems to indicate that ABI breakage in `11.1` was not intentional.

Hence we should probably not re-add `11.1.0` to `11.1.3` to conda-forge.

beckermr:
If they intend to fix the issue, then we may not want to move the pinning into the minor version part. We could simply mark any 11.1.* fmt broken, rebuild the few packages that built against 11.1* against 11.0* instead, and then wait for the next release to see if things are better.

h-vetinari:
> I updated the comment to indicate the order I would do things in.

I broadly agree with that list, though I would start with this PR to fix the observed problems, and stop more packages from being built with `>=11.1` constraints. And if the upstream breakage indeed was unintentional (as happened in the past), then we wouldn't even need step 1 (i.e. tightening the run-exports), we'd just need to wait for a patch version 1.1.5 that's ABI-compatible with 1.0.x

beckermr:
This PR will leave some environments unsolvable if a version of a downstream package has not been built against fmt 11.0. I leave it to you to either rebuild the downstreams or proceed on tightening the pins+run-exports.

h-vetinari:
Out of the builds that have a `>=11.1` constraint, do we want to repopatch them to `>=11.0`, or mark them as broken? I think the former would be better, and if any other problems occur, we still have the option to do the latter.

minrk:
I think it's probably better to mark the 11.1 dependents as broken or leave them unsolvable for now. Allowing them to install with 11.0 is currently leading to broken envs: https://github.com/conda-forge/dolfinx_mpc-feedstock/issues/27

PR #1455, opened 2025-03-26
Add expected outputs for pandas-openscm

znicholls:
## Checklist:

We release a locked version of our package, that pins all dependencies such that users can (if they want to) install without getting any surprises from updates in downstream repos. We forgot to add this in our initial registration, so would like to add it now. The PR changing the recipe is here: https://github.com/conda-forge/pandas-openscm-feedstock/pull/3/files#diff-f3725a55bf339595bf865fec73bda8ac99f283b0810c205442021f29c06eea9a

Thanks for your help!


znicholls:
ping @conda-forge/pandas-openscm

PR #1463, opened 2025-03-29
archive bcolz

h-vetinari:
Upstream https://github.com/Blosc/bcolz has been archived for almost 5 years and doesn't support anything past py3.8

There's been no response in https://github.com/conda-forge/bcolz-feedstock/issues/20 for 1.5 years, which is also a good indicator that it's time.

### DIFF
--- a/requests/bcolz.yml
+++ b/requests/bcolz.yml
@@ -1,0 +1,3 @@
+action: archive
+feedstocks:
+  - bcolz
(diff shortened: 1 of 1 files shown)


h-vetinari:
The only dependent package on bcolz in zipline, which itself hasn't gotten updates in ~5 years: https://github.com/quantopian/zipline. See https://github.com/conda-forge/zipline-feedstock/issues/3.

Actually both bcolz and zipline seem to have follow-up projects with their own feedstocks, see https://github.com/conda-forge/bcolz-zipline-feedstock & https://github.com/conda-forge/zipline-reloaded-feedstock

PR #1468, opened 2025-04-01
Mark rc builds of arcticdb post 5.1.2 as broken

G-D-Petrov:

## Checklist:

Context
See conda-forge/arcticdb-feedstock#426.

Details for rc

These are just release candidate versions of the affected versions, so it is safe to remove them.


xhochy:
Sames as in #1467 

Also: You should not be uploading RC builds to the main label.

jjerphan:
RC builds are published on `conda-forge/label/arcticdb_rc`, do we need to specify this channel explicitly?

xhochy:
No, for that the mark broken automation doesn't work AFAIK.

h-vetinari:
> No, for that the mark broken automation doesn't work AFAIK.

I can mark them broken manually if you want (that's the only option for non-`main` channels)

jjerphan:
It would be helpful. :pray: 

All the rc builds in `>=5.2.1rc0<5.3` might be affected.

h-vetinari:
Oh wow, I did not see the number of builds involved 🤯 

This is a non-batch process (needs to be done manually in UI), so I'm not going to click through 200+ packages. If someone has (or creates) a python script using `anaconda-client` that works (i.e. adds a broken label to a package in an existing channel), I'm happy to run that.

For now my suggestion would be to just rebuild all maintenance branches, and rely on the solver picking up the newest builds.

I don't understand though why you need to draw such a wide circle (rc-channel aside) - we tightened the run-exports of all packages on main to `<11.1a0` (https://github.com/conda-forge/conda-forge-repodata-patches-feedstock/pull/984; at least on windows, we can extend this to other platforms if necessary), so none of the old arcticdb builds should pull in `fmt 11.1` anymore.

jjerphan:
> I don't understand though why you need to draw such a wide circle (rc-channel aside) - we tightened the run-exports of all packages on main to <11.1a0 (https://github.com/conda-forge/conda-forge-repodata-patches-feedstock/pull/984; at least on windows, we can extend this to other platforms if necessary), so none of the old arcticdb builds should pull in fmt 11.1 anymore.

OK, I thought this was done for all platforms but it only has been the case for Windows, but users reported problems on Linux as well.

Could we also apply this repodata patch to `linux-*` and `osx-*`?

h-vetinari:
> Could we also apply this repodata patch to `linux-*` and `osx-*`?

Please open two separate PRs that include osx in the `subdir` filter for the yaml in https://github.com/conda-forge/conda-forge-repodata-patches-feedstock/pull/984, resp. remove the filter entirely for the second PR. We need to do this in stages because we cannot patch ~10k packages at once.

h-vetinari:
Two PRs. One that extends the subdir condition for the yaml in https://github.com/conda-forge/conda-forge-repodata-patches-feedstock/pull/984 from windows to windows+osx, and then a second one that enables the patch on linux too by deleting the subdir condition. You can also switch the order of linux and osx if you want.

Repodata patches are against the current state of the repodata. Windows is already patched (hence why we keep patching it), now we need to add osx and linux in two waves, because there's a practical limit of ~5000 packages being patched in one go (second round a few hours later)

jjerphan:
OK, this is much clearer, thank you.

carterbox:
My understanding is that these packages were published to an RC channel (not the main channel) and thus this admin request will not actually mark anything as broken. Thus, there is nothing to do here.

PR #1495, opened 2025-04-17
fix: move staged-recipes creation scripts to admin-requests

beckermr:
This PR moves the staged-recipes creation scripts to admin requests to help further ensure they do not get changes. We have human review and the linter, but this will prevent any accidents or things missed on CR from causing issues.


beckermr:
@jaimergp This one is ready to go!

PR #1501, opened 2025-04-24
Allow multiple outputs that start with torchcodec- in the torchcodec feedstock

traversaro:
In https://github.com/conda-forge/torchcodec-feedstock/pull/5 I added a torchcodec-tests output to the torchcodec feedstock, as originally requested in https://github.com/conda-forge/staged-recipes/pull/29447#discussion_r2019978513 .

To handle this request and possible similar modifications in the future, I think it could make sense to allow all the outputs with `torchcodec-` prefix to `torchcodec` feedstock, as anyhow the prefix is not ambigous. If instead you prefer that I explicitly list the requested outputs (now `torchcoded`, and in the future possibly more) I can listed them explicitly, thanks!

## Checklist:

ping @conda-forge/torchcodec (that is actually just me)

### DIFF
--- a/requests/torchcodec.yaml
+++ b/requests/torchcodec.yaml
@@ -1,0 +1,3 @@
+action: add_feedstock_output
+feedstock_to_output_mapping:
+  - torchcodec: "torchcodec-*"
--- a/requests/torchcodec.yaml
+++ b/requests/torchcodec.yaml
@@ -1,3 +1,3 @@
 action: add_feedstock_output
 feedstock_to_output_mapping:
-  - torchcodec: "torchcodec-*"
+  - torchcodec: "torchcodec-tests"


requests/torchcodec.yaml
```diff
@@ -0,0 +1,3 @@
+action: add_feedstock_output
+feedstock_to_output_mapping:
+  - torchcodec: "torchcodec-*"
```
hmaarrfk, line 3:
```suggestion
  - torchcodec: "torchcodec-tests"
```

I personally don't like globbing. is this acceptable?

traversaro: Sure! That was indeed mentioned in the original post, even if there were a few copy&paste error for which the message was confusing.

PR #1504, opened 2025-04-28
Reset token for rustac-feedstock

weiji14:
Accidentally mixed v0 `meta.yaml` and v1 `recipe.yaml` syntax at https://github.com/conda-forge/staged-recipes/pull/29772 (see https://github.com/conda-forge/rustac-feedstock/pull/1#issuecomment-2836474587), and builds are failing with this error:

```
validation results:
{
  "osx-64/rustac-0.6.0-py312h5cd10b2_0.conda": false
}
NOTE: Any outputs marked as False are not allowed for this feedstock. See https://conda-forge.org/docs/maintainer/infrastructure/#output-validation-and-feedstock-tokens for information on how to address this error.
```

hopefully a token reset can fix this, and we can merge https://github.com/conda-forge/rustac-feedstock/pull/2. Cc @gadomski.

## Checklist:

### DIFF
--- a/requests/rustac-token-reset.yml
+++ b/requests/rustac-token-reset.yml
@@ -1,0 +1,3 @@
+action: token_reset
+feedstocks:
+- rustac


xhochy:
You will need https://github.com/conda-forge/admin-requests/blob/main/examples/example-add-feedstock-output.yml instead

weiji14:
Thanks @xhochy, wasn't sure which request to put in. Have changed things to the `-add-feedstock-output.yml` now.

requests/rustac-add-feedstock-output.yml
```diff
@@ -0,0 +1,3 @@
+action: add_feedstock_output
+feedstock_to_output_mapping:
+  - rustac: rustac
```
weiji14, line 3: @gadomski, just to double check, you want the package name for rustac's Python bindings to be just `rustac` and not `rustac-py`? Do you want to have R bindings `rustac-r` as mentioned at https://github.com/stac-utils/rustac/issues/641 in the future? And potentially pure Rust `rustac` as well?

gadomski: >  you want the package name for rustac's Python bindings on conda-forge to be just rustac and not rustac-py

Yeah, I think this is correct ... I'm leaning in to letting Python and Rust share the **rustac** name. Since https://github.com/stac-utils/rustac-py includes a wrapper around the Rust CLI, we shouldn't need the Rust package on Conda. If we make an R one, we'll have to use **rustac-r** 🤷🏼.

xhochy: It will be `r-rustac`, we prefix all R packages on conda-forge.

weiji14: Ok, then should be all good to go :rocket:

PR #1511, opened 2025-05-05
Add outputs for `apache-airflow-providers-amazon-with-*` extras

xylar:
This is needed for https://github.com/conda-forge/apache-airflow-providers-amazon-feedstock/pull/69, where we are hoping to add the extras.  The extras are, in turn, needed for a new core apache-airflow package being added in https://github.com/conda-forge/staged-recipes/pull/29879.

## Checklist:

ping @conda-forge/apache-airflow-providers-amazon


requests/add-apache-airflow-providers-amazon-outputs.yml
```diff
@@ -0,0 +1,3 @@
+action: add_feedstock_output
+feedstock_to_output_mapping:
+  - apache-airflow-providers-amazon: "apache-airflow-providers-amazon-with-*"
```
hmaarrfk, line 3: i personally won't merge globs. please be specific.

xylar: @hmaarrfk, oh, interesting.  I haven't got push-back on this before so I assumed it was an allowable practice.  I'll change it.

It does mean that each time a new extra gets added, it means another admin request for airflow-related feedstocks, that's a lot of extra maintenance work.  There are already 70 or 80 provider repos, which is a lot of maintenance burden on me.  Fortunately, this is the only one so far with extras that are actually required by other feedstocks but this could quickly balloon!

But we can revisit this if it becomes an issue.

hmaarrfk: yeah. i guess different core members might be following different guidelines. But to me, globs just ask for trouble in the long run.

Better to be specific, and deliberate.

xylar: And I guess you don't mind the extra traffic here as new extras get added?

xylar:
@hmaarrfk, I think I've addressed your comments.  Do you have time to give it another look?

xylar:
@hmaarrfk, thanks so much!

PR #1516, opened 2025-05-07
Request larger runners for VTK-m

vicentebolea:
## Checklist:

@conda-forge/vtk-m

VTK-m is toolkit for scientific visualization that heavily uses devices adapters such as GPUs to speedup its algorithms. VTK-m build times are very extended due to the heavy use of  metaprogramming in its source code, this is specially true when using CUDA which eventually yields builds exceeding the 6 hours max build time in conda-forge standard CI resources.

I have a PR adding CUDA to VTK-m, however, I cannot merge it since the build times sometime exceeds our 6h allocation. Having more or faster resources would alleviate this.

https://github.com/conda-forge/vtk-m-feedstock/pull/9

We need VTK-m to support CUDA in Conda since it is the only way for VTK and ParaView to support CUDA for its algorithms, since they delegate in VTK-m for GPU accelerated algorithms. This will greatly benefits the Scientific visualization community.

### DIFF
--- a/requests/vtk-m-cirun-request.yml
+++ b/requests/vtk-m-cirun-request.yml
@@ -1,0 +1,15 @@
+action: cirun
+feedstocks:
+  # list of feedstock names (sans `-feedstock` suffix)
+  - vtk-m
+resources:
+  # list of resource names you are applying to
+  # see https://github.com/conda-forge/.cirun for available options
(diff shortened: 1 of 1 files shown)


vicentebolea:
How can I get some eyes on this? Thanks

PR #1529, opened 2025-05-20
Add 2 more `airflow-with-*` outputs

xylar:
Aargh!  I somehow missed 2 more outputs in in https://github.com/conda-forge/admin-requests/pull/1526 and https://github.com/conda-forge/admin-requests/pull/1527.  These are the last ones for now, I promise!

ping @conda-forge/airflow

Needed for https://github.com/conda-forge/airflow-feedstock/pull/148

## Checklist:

### DIFF
--- a/requests/add-airflow-outputs.yml
+++ b/requests/add-airflow-outputs.yml
@@ -1,0 +1,4 @@
+action: add_feedstock_output
+feedstock_to_output_mapping:
+  - airflow: airflow-with-apache-beam
+  - airflow: airflow-with-neo4j


xylar:
@conda-forge/core, this is ready for review.  I'm really sorry for the sloppiness.  I though I had it the last time but I really think this is it.

xhochy (approved):

xylar:
Thanks again @xhochy!  Sorry for wasting more of your time.

xhochy:
You could also request a wildcard next time, then you don't need to do so many individual whitelists.

xylar:
@xhochy, I used to do that and got some grief, see https://github.com/conda-forge/admin-requests/pull/1511#discussion_r2073478550.  As discussed there, there seems to be disagreement among folks in `core` about whether wildcards are "allowed" or not.  I sure would prefer to use them but I didn't want more pushback.

hmaarrfk:
yeah, sorry thats me, no liking wildcards.

PR #1554, opened 2025-06-09
Discourage the use of glob patterns when requesting additional outputs

hmaarrfk:
I personally think that we shouldn't use globs ever. They squat on future outputs and can have "unbound" consequences.

Instead, I want to push people to use very specific output names.

Essentially, I want to decrease my mental load when reviewing these PRs.

With the glob, it becomes a "discussion". without the glob, it is SUPER easy.

### DIFF
--- a/README.md
+++ b/README.md
@@ -79,5 +79,11 @@ By default, `conda-forge` feedstocks cannot push packages to our channel that an
 when building your package indicating that the given package was not allowed for your feedstock (e.g., you moved a package
 build from one feedstock to another), you should request the output be added to the new feedstock via this repository. An example request
 is located in [examples/example-add-feedstock-output.yml](examples/example-add-feedstock-output.yml). You can add both glob patterns
-and package names. We support the glob syntax of the Python `fnmatch` module. Make a PR putting your
-`.yml` request file in the `requests` directory and the `conda-forge/core` team will review it.
+and package names.
+
+While glob patterns are support, they should be used with care as they
+essentially "squat" on all future matched. If you are requesting a specific
+package output, please use the full name of the package output. We support the
+glob syntax of the Python `fnmatch` module. Make a PR putting your `.yml`
+request file in the `requests` directory and the `conda-forge/core` team will
+review it.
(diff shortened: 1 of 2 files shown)


examples/example-add-feedstock-output.yml
```diff
@@ -4,5 +4,6 @@ feedstock_to_output_mapping:
   - clang-compiler-activation: clang_impl_osx-64
   # this entry adds another output to the same feedstock
   - clang-compiler-activation: clang_impl_osx-arm64
+  # While fully specified package names are preferred, you can also use
   # this entry adds an allowed glob pattern
```
xylar, line 8:
```suggestion
  # this entry to add an allowed glob pattern
```

hmaarrfk:
thanks all

README.md
```diff
@@ -79,5 +79,11 @@ By default, `conda-forge` feedstocks cannot push packages to our channel that an
 when building your package indicating that the given package was not allowed for your feedstock (e.g., you moved a package
 build from one feedstock to another), you should request the output be added to the new feedstock via this repository. An example request
 is located in [examples/example-add-feedstock-output.yml](examples/example-add-feedstock-output.yml). You can add both glob patterns
-and package names. We support the glob syntax of the Python `fnmatch` module. Make a PR putting your
-`.yml` request file in the `requests` directory and the `conda-forge/core` team will review it.
+and package names.
+
+While glob patterns are support, they should be used with care as they
+essentially "squat" on all future matched. If you are requesting a specific
```
jakirkham, line 85: I would be interested to know more what you mean by "squat"?

Asking because it is possible for two feedstocks to upload under the same name if both are allowed. We have done this when we migrate a feedstock and package name from `_` to `-` to match PyPI

hmaarrfk: For example, just because my package is called `numpy`, why should I get to have the output of every single package with `numpy-*` out there.

pypi shows many packages, from people that use `numpy-` in their name at the front.

https://pypi.org/search/?q=numpy


If we want people to be explicit, then just let that be the case, and let us do the review quickly. 

If i allow `numpy-*` then I need to be vigilent when i merge in future `numpy-enhancement` from `staged-recipes` for fear of collisions.

This stuff isn't needed. If we want to create the ability to have name space, which I think are a good idea, we should create this feature in conda upstream.

But for now, I think that as long as we have a single global namespace, its alot faster to review things when people list out what they want, then my review includes searching for a few items on https://conda-forge.org/packages/ and pressing merge. instead of trying to think of the unbound downstream consequences.

PR #1556, opened 2025-06-10
Mark fastjet-contrib-static v1.101 build 2 as broken

matthewfeickert:
* fastjet-contrib-static v1.101 build 2 had dynamic libraries added to it including a version that is known to cause problems, which is why the -static version was created. These should not be merged, and so build 2 needs should be considered broken.
   - c.f. https://github.com/conda-forge/fastjet-contrib-feedstock/pull/9

## Checklist:

  * [x] Pinged the team for the package for their input. (I am part of @conda-forge/fastjet-contrib)


lgray:
The version that can cause problems is not installed in `-static`. Please correct that assertion.

There is no longer the (straightforward) possibility of building the static libraries in any case, it is factored between `fastjet-contrib` and `fastjet-contrib-static` in terms of package contents as it was before when using the cmake setup. The fragile library is not even built for the `-static` output case, only the individual libraries are.

We should likely instead rename `-static` to `-split`, or something like that, and deprecate `-static`.

Concerning process here, I suppose it's misleading rather than broken, but likely worth it to mark it as such.

matthewfeickert:
> The version that can cause problems is not installed in `-static`. Please correct that assertion.

You're right (for others c.f. https://github.com/conda-forge/fastjet-contrib-feedstock/pull/9#issuecomment-2960524767). PR body updated.

> There is no longer the (straightforward) possibility of building the static libraries in any case, it is factored between `fastjet-contrib` and `fastjet-contrib-static` in terms of package contents as it was before when using the cmake setup. The fragile library is not even built for the `-static` output case, only the individual libraries are.
> 
> We should likely instead rename `-static` to `-split`, or something like that, and deprecate `-static`.
> 
> Concerning process here, I suppose it's misleading rather than broken, but likely worth it to mark it as such.

conda-forge doesn't have more nuance in its labels, so "broken" gets used to label many things where the build is not "broken" but contains outputs that it shouldn't, like in this case. The https://github.com/conda-forge/fastjet-contrib-feedstock/ outputs will need to change given https://github.com/conda-forge/fastjet-contrib-feedstock/pull/9, which will deprecate `-static` and need to create new output names to differentiate between the dynamic libraries split out across multiple libraries, and the "old style" of having the "fragile" single dynamic library where everything gets merged. Though @lgray, given that you've solved the "fragile" nature here it sounds like it might be possible to just go down to a single output with dynamic libraries, if there's nothing about the ABI or end user API that changes. :+1:

PR #1567, opened 2025-06-19
Add outputs for jupytercad

arjxn-py:
This is to allow https://github.com/conda-forge/jupytercad-feedstock/issues/23

### DIFF
--- a/requests/jupytercad-add-outputs.yml
+++ b/requests/jupytercad-add-outputs.yml
@@ -1,3 +1,11 @@
 action: add_feedstock_output
 feedstock_to_output_mapping:
-  - jupytercad: "jupytercad*"
+  - jupytercad: "jupytercad"
+  - jupytercad: "jupytercad-core"
+  - jupytercad: "jupytercad_core"
+  - jupytercad: "jupytercad-lab"
+  - jupytercad: "jupytercad_lab"
+  - jupytercad: "jupytercad-lite"
+  - jupytercad: "jupytercad_lite"
+  - jupytercad: "jupytercad-app"
+  - jupytercad: "jupytercad_app"
--- a/requests/jupytercad-add-outputs.yml
+++ b/requests/jupytercad-add-outputs.yml
@@ -1,0 +1,3 @@
+action: add_feedstock_output
+feedstock_to_output_mapping:
+  - jupytercad-packages: "jupytercad*"
--- a/requests/jupytercad-add-outputs.yml
+++ b/requests/jupytercad-add-outputs.yml
@@ -1,11 +1,6 @@
 action: add_feedstock_output
 feedstock_to_output_mapping:
-  - jupytercad: "jupytercad"
   - jupytercad: "jupytercad-core"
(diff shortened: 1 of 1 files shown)


SylvainCorlay (approved):

SylvainCorlay:
This looks sensible, but I have never worked on the output mapping.

h-vetinari:
Glob patterns should be used sparingly (c.f. #1554). It's better to be explicit, e.g. just add `jupytercad-core` or whatever handful of names you want to add. If you use the glob, and sometimes later you (or someone else!) end up needing another feedstock `jupytercad-foo` (e.g. some other component or downstream use-case), the glob will turn out to be wrong.

requests/jupytercad-add-outputs.yml
```diff
@@ -0,0 +1,3 @@
+action: add_feedstock_output
+feedstock_to_output_mapping:
+  - jupytercad: "jupytercad*"
```
martinRenou, line 3:
```suggestion
  - jupytercad: "jupytercad"
  - jupytercad: "jupytercad-core"
  - jupytercad: "jupytercad_core"
  - jupytercad: "jupytercad-lab"
  - jupytercad: "jupytercad_lab"
  - jupytercad: "jupytercad-lite"
  - jupytercad: "jupytercad_lite"
  - jupytercad: "jupytercad-app"
  - jupytercad: "jupytercad_app"
```

requests/jupytercad-add-outputs.yml
```diff
@@ -0,0 +1,11 @@
+action: add_feedstock_output
+feedstock_to_output_mapping:
+  - jupytercad: "jupytercad"
+  - jupytercad: "jupytercad-core"
+  - jupytercad: "jupytercad_core"
```
hmaarrfk, line 5: are these underscores really necessary? with this kind of "output validation" you can "start off" with the "correct one".

They somewhat pollute environments and cause non-ssds to seriously slow down.

hmaarrfk: oh ok. yeah you don't need to "add" these here since they are already "outputs".

You only need to add 'new outputs'

requests/jupytercad-add-outputs.yml
```diff
@@ -0,0 +1,11 @@
...
+  - jupytercad: "jupytercad-lab"
+  - jupytercad: "jupytercad_lab"
+  - jupytercad: "jupytercad-lite"
+  - jupytercad: "jupytercad_lite"
+  - jupytercad: "jupytercad-app"
+  - jupytercad: "jupytercad_app"
```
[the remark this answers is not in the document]
martinRenou: Thanks!

h-vetinari:
> @h-vetinari would you kindly review again?

Sorry for the delay, I'm listening to \*all\* notifications from this repo, but in turn that sometimes mean I don't see thing directly addressed to me right away.

martinRenou:
No worries at all, thanks for reviewing!

PR #1568, opened 2025-06-19
Mark `patchelf` version `0.18.0` broken

baszalmstra:
New versions of `patchelf` `0.18` were released when the feedstock moved to recipe v1. However, the `0.18` versions should be marked as broken: https://github.com/conda-forge/admin-requests/pull/746

@xhochy @h-vetinari 

---

## Checklist:

### DIFF
--- a/requests/patchelf018.yaml
+++ b/requests/patchelf018.yaml
@@ -1,0 +1,9 @@
+action: broken
+packages:
+- linux-64/patchelf-0.18.0-h3f2d84a_2.conda
+- osx-64/patchelf-0.18.0-h92383a6_2.conda
+- linux-aarch64/patchelf-0.18.0-h5ad3122_2.conda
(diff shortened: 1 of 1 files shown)


h-vetinari:
As I said on the zulip thread, if @xhochy et al need v0.18, we should simply change the label in the [anaconda UI](https://anaconda.org/conda-forge/patchelf/files), which is a manual action in any case.

Something like `patchelf_dev` / `patchelf_next` or whatever

h-vetinari:
I did this manually now

image

#854 didn't touch the osx builds for `_1`, and I left that unchanged; OTOH it seemed more consistent to me to put all the `_2` builds under a consistent label, because that way we can update the feedstock to match it, so that we avoid the issue of accidentally pushing to main again.

h-vetinari (changes requested):
has become obsolete

PR #1589, opened 2025-07-23
archive librmm

h-vetinari:
See https://github.com/conda-forge/librmm-feedstock/issues/75

### DIFF
--- a/requests/librmm.yml
+++ b/requests/librmm.yml
@@ -1,3 +1,4 @@
 action: archive
 feedstocks:
   - librmm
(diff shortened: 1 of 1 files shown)


leofang (approved):

h-vetinari:
Just noticed that the rmm feedstock depends on librmm. It seems to have fallen into the same kind of disrepair as librmm, so I'm adding it for archival here as well. Xref https://github.com/conda-forge/rmm-feedstock/issues/30

PR #1592, opened 2025-07-24
Request to add output: ufs2arco-nompi

timothyas:
Hello! I would like to add an optional build to the package [ufs2arco](https://anaconda.org/conda-forge/ufs2arco) which does not install mpi, so users can `conda install -c conda-forge ufs2arco-nompi`. I am very new to this, but I'm hoping this is the right way to request this additional output.

I am currently the owner/maintainer of the repo/package.

## Checklist:

### DIFF
--- a/requests/ufs2arco-add-feedstock-output.yml
+++ b/requests/ufs2arco-add-feedstock-output.yml
@@ -1,0 +1,3 @@
+action: add_feedstock_output
+feedstock_to_output_mapping:
+  - ufs2arco: ufs2arco-nompi


xhochy:
Please link the PR here where you add the output.

timothyas:
Thanks @hmaarrfk for the suggestion, I think this is what I would rather do too. Am I understanding this correctly, that hdf5 uses if statements instead of having multiple outputs to handle different mpi (or nompi) situations? 

I attempted this in my latest commit to [this ufs2arco-feedstock PR](https://github.com/conda-forge/ufs2arco-feedstock/pull/2). Note that I don't really want to handle every different mpi case, since I'm really piggy backing MPI support through mpi4py, so I don't think it's necessary.

timothyas:
I'll go ahead and close this. @hmaarrfk helped me a ton with getting different builds rather than different outputs, so this is no longer necessary. Thank you!

PR #1600, opened 2025-08-02
Make the `create_feedstocks` script compatible with the PyGitHub 2.7.0 API

shermansiu:
- The current `create_feedstocks` actions keep crashing (e.g. https://github.com/conda-forge/admin-requests/actions/runs/16696874176)
- The PyGitHub repo recently changed up their API in a backwards-incompatible way, which led to this problem
    - https://github.com/PyGithub/PyGithub/releases/tag/v2.7.0
- The `create_feedstocks` action started crashing 2 days ago, which is exactly when the v2.7.0 PyGitHub version got released.

## Other potential fixes
- `verification` does not appear in this repo, so we don't need to fix that (https://github.com/search?q=repo%3Aconda-forge%2Fadmin-requests%20verification&type=code)
- The only appearance of `get_rate_limit` in this repo is in `scripts/create_feedstocks.py`, so this is the only part of the repo that needs to be fixed (https://github.com/search?q=repo%3Aconda-forge%2Fadmin-requests%20get_rate_limit&type=code)

### DIFF
--- a/scripts/create_feedstocks.py
+++ b/scripts/create_feedstocks.py
@@ -204,11 +204,11 @@ def print_rate_limiting_info(gh, user):
     # spending it and how to better optimize it.
 
     # Get GitHub API Rate Limit usage and total
-    gh_api_remaining = gh.get_rate_limit().core.remaining
-    gh_api_total = gh.get_rate_limit().core.limit
+    gh_api_remaining = gh.get_rate_limit().resources.core.remaining
+    gh_api_total = gh.get_rate_limit().resources.core.limit
 
     # Compute time until GitHub API Rate Limit reset
-    gh_api_reset_time = gh.get_rate_limit().core.reset
+    gh_api_reset_time = gh.get_rate_limit().resources.core.reset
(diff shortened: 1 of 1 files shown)


h-vetinari:
Thanks for digging into this. For now I think I'd prefer to add `<2.7` to 
https://github.com/conda-forge/admin-requests/blob/80e2b9b9693e45407a4e9975ec795bbbc34f8fce/scripts/create_feedstocks#L48
(also in `environment.yaml`) and then later do an explicit changeover to the new API (who needs semantic versioning anyway? 🤦‍♂️) while raising the lower bound

hmaarrfk:
Or at least they could havea  `__version__` variable like most sane projects....
https://github.com/PyGithub/PyGithub/issues/3325 <-- related question

hmaarrfk:
@h-vetinari ill try to make the Pr to limit pygithub. 

But would you be ok with us using hasattr to write compatibility shims for both pre 2.7 and 2.7. 

At least for the next 6 months while the API stabilizes.

PR #1635, opened 2025-09-01
mark llvm-openmp v21.1.0 broken on windows

h-vetinari:
Apparently, openmp v21 breaks numpy on windows: https://github.com/conda-forge/numpy-feedstock/issues/365

It was released just ~12h ago; mark it as broken to stop the bleeding while we investigate.

CC @conda-forge/numpy @conda-forge/scipy @conda-forge/core

### DIFF
--- a/requests/openmp.yml
+++ b/requests/openmp.yml
@@ -1,0 +1,3 @@
+action: broken
+packages:
+-  win-64/llvm-openmp-21.1.0-hfa2b4ca_0.conda


rouault:
This also broke GDAL on Windows, when using libarmadillo, which uses blas underneath.

h-vetinari:
On windows, or also on other platforms? For windows, you should get it in newly installed/updated environments anymore, since the build was marked as broken

rouault:
> On windows

Windows only apparently. I confirm that the broken package is no longer installed when creating a new env . Thanks

PR #1664, opened 2025-09-16
Mark sagemaker code editor feedstock with 1.8.0b1,1.8.0b2,1.8.0b3,1.8.0b4,1.8.0b5,1.8.0.dev0 as broken

aakashmandavilli96:

## Checklist:

* [x] I want to mark a package as broken :

- We have pushed our dev builds and beta builds to the main folder and these builds should be marked as broken as this is impacting multiple users 
- Following builds from `1.8.0b1 to 1.8.0b5` should be marked as broken along with the `1.8.0.dev0`


aakashmandavilli96:
@conda-forge-admin can we get this reviewed and merged

PR #1687, opened 2025-10-03
Remove cuDNN CUDA12 packages which are missing device code for pre-Turing

carterbox:
Any cuDNN package for cuDNN>=9.11 does not support pre-Turing devices. However, pre-Turing devices are still supported by CUDA until CUDA 13. Since we cannot check the devices on a users system. We instead mark these packages as broken for anyone running CUDA 12. Turing and later users can upgrade to CUDA 13 to get the latest version of cuDNN.

https://github.com/conda-forge/cudnn-feedstock/issues/124

## Checklist:

### DIFF
--- a/requests/cudnn-broken.yml
+++ b/requests/cudnn-broken.yml
@@ -1,0 +1,18 @@
+action: broken
+packages:
+# libcudnn packages missing device code for pre-Turing, but built for CUDA 12
+  - linux-64/libcudnn-9.11.0.98-hf7e9902_0.conda
+  - linux-aarch64/libcudnn-9.11.0.98-h703c024_0.conda
+  - win-64/libcudnn-9.11.0.98-hca898b4_0.conda
+
+  - linux-64/libcudnn-9.12.0.46-hf7e9902_0.conda
+  - linux-aarch64/libcudnn-9.12.0.46-hc13e719_0.conda
+  - win-64/libcudnn-9.12.0.46-hca898b4_0.conda
+
+  - linux-64/libcudnn-9.13.0.50-hf7e9902_0.conda
+  - linux-aarch64/libcudnn-9.13.0.50-hc13e719_0.conda
+  - win-64/libcudnn-9.13.0.50-hca898b4_0.conda
+
+  - linux-64/libcudnn-9.13.1.26-hf7e9902_0.conda
+  - linux-aarch64/libcudnn-9.13.1.26-hc13e719_0.conda
+  - win-64/libcudnn-9.13.1.26-hca898b4_0.conda


h-vetinari (commented):
Assuming we actually want to do this, then this needs to cover all the cudnn component packages that depend on libcudnn, at least `cudnn` and `libcudnn-dev` (not sure how the `-jit` variants interact, but presumably those too). Otherwise the solver will find a "valid" `cudnn` build and then complain about an unresolvable dependency (the `libcudnn` marked broken).

More importantly, we'll have to think about patching out constraints that force these patches, see https://github.com/conda-forge/conda-forge-repodata-patches-feedstock/pull/1090. Luckily, there may not be much or any of those.

carterbox:
> (not sure how the -jit variants interact, but presumably those too).

Yes. Thanks for catching this. I do need to include `libcudnn-jit`.

>  Otherwise the solver will find a "valid" cudnn build and then complain about an unresolvable dependency (the libcudnn marked broken).

There are exact pinnings between the dev and runtime components of cudnn, so this type of failure is fine.

carterbox:
We may need to try to backport the `cudnn-jit` metapackage creation to `cudnn==9.10.x`.

kmittman:
> We may need to try to backport the `cudnn-jit` metapackage creation to `cudnn==9.10.x`.

I believe that's missing by design, the full roll-out was 9.12.x

h-vetinari (approved):
Thanks

hmaarrfk:
Can we please undo this.

Its causing tensorflow to be unusable
https://github.com/conda-forge/conda-forge-repodata-patches-feedstock/pull/1090#issuecomment-3440728009

hmaarrfk:
I think that an other solution must be found.

I think it is commendable that you tried to fix things for one subset of users (Ampere GPU users 10series, i have a few of those in my lab), but the truth is that this really hurts users of newer cards.

This broke tensorflow all the way back to 2.18 which was released in Feb 2025.

So 8 months of tensorflow (2.18 + 2.19) is now unusable?

I would rather follow this approach:

1. This patch, while trying to help one subset of users has hurt an other. And thus should be reverted.
2. Nvidia has worked to better expose its capabilities to conda and we should work with them to release https://github.com/conda-forge/conda-forge.github.io/issues/2623
3. Once the patch above is released we should use it to better repo-data patch and not just mark all these working for "turing+ users" packages.
4. Temporarily, we should let users of turning GPU know that they should stick to cudnn 9.10, and add that in their requirements.

cc: @conda-forge/tensorflow

traversaro:
While I agree with @h-vetinari in theory, in practice the new situation of CUDA-powered "jax and tensorflow" are broken is much worse (and clearly that is a subjective evaluation) that then previous situation of cudnn is broken on some old GPUs. I will be happy in helping rebuild the world with cudnn=9.10.*, but while we work on that it would be great if we (temporarly) mark the cudnn builds as not broken again as a hot fix.

PR #778, opened 2023-07-31
archive weave-feedstock

mstimberg:
## Checklist:

* [x] I want to archive a feedstock:
  * [x] Pinged the team for that feedstock for their input.
  * [x] Make sure you have opened an issue on the feedstock explaining why it was archived.
  * [x] Linked that issue in this PR description.
  * [x] Added links to any other relevant issues/PRs in the PR description.

Pinging @conda-forge/weave (but I am the only maintainer on the package).

Issue at `weave-feedstock`: conda-forge/weave-feedstock#21

For explanation and rationale, see conda-forge/admin-requests#766

Short summary: The upstream package has been renamed to scipy-weave, but is basically frozen and will not be updated to work with recent Python/numpy versions. The package that is now named [weave on PyPi](https://pypi.org/project/weave/) is unrelated.

PR #1711, opened 2025-10-21
Un-archive `weave` feedstock

andrewtruong:
Hi, I'd like to un-archive the `weave` feedstock and make me the new maintainer.

The old project called `weave` was archived per https://github.com/conda-forge/admin-requests/pull/778.  The new project, now called `weave` on pypi, would like to start publishing on conda.

### DIFF
--- a/requests/unarchive-weave-feedstock.yml
+++ b/requests/unarchive-weave-feedstock.yml
@@ -1,0 +1,3 @@
+action: unarchive
+feedstocks:
+  - weave


beckermr (changes requested):
It appears that you'd like to point the same name to a completely different package? We do not allow that on conda-forge.

andrewtruong:
I understand your concern.

The pypi package name changed over 2 years ago, and the conda package has not had any updates in almost 6 years.  I think we can draw a clean line separating the old vs. new packages by version number -- namely anything `weave>=0.20.0` is the new package, and below is the old package (the latest version of the old package is `0.17.0`)

jaimergp:
xref https://github.com/conda-forge/cfep/pull/64

carterbox:
@andrewtruong, [according the CFEP-26](https://github.com/conda-forge/cfep/blob/main/cfep-26.md#names-in-use), you will need to mark all of the existing weave packages as broken instead of unarchiving. Then my understanding is that you will make a new feedstock via staged-recipes with a different feedstock name.

PR #1769, opened 2025-11-26
Add spheno-atlas output for ATLAS experiment variations

matthewfeickert:

## Checklist:

     - I am the maintainer @conda-forge/spheno
     - In https://github.com/conda-forge/spheno-feedstock/pull/3 there are patches that are needed for the ATLAS experiment that enable options that change numerical output in the tests. To confine these changes to ATLAS only, https://github.com/conda-forge/spheno-feedstock/pull/3 adds the `spheno-atlas` output given recommendations in [#general > Easy way to see all build variants available to users?](https://conda-forge.zulipchat.com/#narrow/channel/457337-general/topic/Easy.20way.20to.20see.20all.20build.20variants.20available.20to.20users.3F/with/558854976) to avoid users manually requesting build variants.

### DIFF
--- a/requests/add-spheno-atlas-output.yml
+++ b/requests/add-spheno-atlas-output.yml
@@ -1,0 +1,3 @@
+action: add_feedstock_output
+feedstock_to_output_mapping:
+  - spheno: spheno-atlas


matthewfeickert:
> Commented on the feedstock, I guess you want this output as a simple way to select the correct build, but not as a general package (alternative). Let's come to a conclusion there and then we can merge here.

Thanks @xhochy. I think given your feedback my plan is now https://github.com/conda-forge/spheno-feedstock/pull/3#issuecomment-3583346896, and as that would not require a new output of the `spheno` feedstock (just a build variant) then this PR is no longer required. If you agree we can close this.

xhochy:
Yes, this sounds sensible.

matthewfeickert:
> However, it'd just be much easier to open a new PR.

Sounds good and will do. :+1:

PR #1784, opened 2025-12-08
Request larger runners with GPU for viskores

vicentebolea:
## Checklist:

@conda-forge/viskores

Viskores (previously known as VTK-m) is a toolkit for scientific visualization that heavily uses devices adapters such as GPUs to speedup its algorithms. Viskores build times are very extended due to the heavy use of  metaprogramming in its source code, this is specially true when using CUDA which eventually yields builds exceeding the 6 hours max build time in conda-forge standard CI resources.

I have a PR adding CUDA to Viskores, however, I cannot merge it since the build times sometime exceeds our 6h allocation. Having more or faster resources would alleviate this.

https://github.com/conda-forge/viskores-feedstock/pull/6

We need Viskores to support CUDA in Conda since it is the only way for VTK and ParaView to support CUDA for its algorithms, since they delegate in Viskores for GPU accelerated algorithms. This will greatly benefits the Scientific visualization community.

The reason of requesting a GPU runner is that I we need want to smoke test the GPU backend after building.

### DIFF
--- a/requests/viskores-cirun-request.yml
+++ b/requests/viskores-cirun-request.yml
@@ -1,0 +1,14 @@
+action: cirun
+feedstocks:
+  # list of feedstock names (sans `-feedstock` suffix)
+  - viskores
+resources:
+  # list of resource names you are applying to
+  # see https://github.com/conda-forge/.cirun for available options
+  - cirun-openstack-gpu-2xlarge
+# switch to true if you need it on pull-requests too
+pull_request: true  
+# revoke access to the gpu runner
+revoke: false
+# Send an automatic PR to the feedstock to enable cirun defaults
+send_pr: true


vicentebolea:
@h-vetinari when I re-render locally I get: `RuntimeError: Using github_actions as the CI provider inside conda-forge github org is not allowed in order to avoid a denial of service for other infrastructure.
`

h-vetinari:
Ah, sorry, you need to also add
```
github_actions:
  self_hosted: true
```
I just tested this locally and then it works. Also, please consider alphabetizing your existing entries in `conda-forge.yml` in a separate preparatory commit.

h-vetinari:
Are you offering to sponsor another server? :)

There's one big machine. It's shared. Sometimes that means waiting. That's generally still better than not being able to run CI at all.

Depending on how many vCPUs are requested, it can do up to ~10 parallel jobs. You chose the rarest resource (GPU-enabled), of which there are only 4 in parallel (in addition to any limitations caused by CPU consumption), so you'll wait even longer on average. E.g. when the pytorch feedstock does a build, you won't get barely any resources for ~24h.

In turn that means being very judicious with resources while debugging (e.g. skip all but one variants, then unskip for one full CI run before merging). That's on everyone who's using that resource.

vicentebolea:
@h-vetinari find the new PR here: https://github.com/conda-forge/.cirun/pull/135

h-vetinari:
@vicentebolea, I'm reopening this after all; there's a registration step for the new CI provider (per feedstock) that's triggered by this PR.

h-vetinari (approved):

PR #1872, opened 2026-02-06
add regular pre-commit setup

h-vetinari:
Taken from https://github.com/conda-forge/conda-smithy/blob/main/.pre-commit-config.yaml

PR #1874, opened 2026-02-06
add isort to pre-commit

h-vetinari:
Another follow-up to #1872 & #1873. I also noticed that the indentation was very off. The isort config is taken from [conda-forge-metadata](https://github.com/conda-forge/conda-forge-metadata/blob/main/.pre-commit-config.yaml), which has a bunch of other options for hooks that we could consider mirroring as well.

### DIFF
--- a/conda_forge_admin_requests/__init__.py
+++ b/conda_forge_admin_requests/__init__.py
@@ -1,12 +1,13 @@
 import importlib
 import pkgutil
+
 from . import (
-    archive_feedstock,
-    mark_broken,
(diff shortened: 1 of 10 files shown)


beckermr:
I meant both ruff checking and ruff formatting.

```
  - repo: https://github.com/astral-sh/ruff-pre-commit
    rev: v0.14.14
    hooks:
      - id: ruff-check
        args: [ --fix ]
      - id: ruff-format
```

this doesn't actually matter of course. ;p

PR #1879, opened 2026-02-09
Archive nyancad packages

pepijndevos:
NyanCAD has changed direction and is no longer relying on conda-forge for installation.

This should remove an mosaic/pyttoreque/simserver packages.

If I'm correct the only other package I'm maintainer of is CouchDB and possibly any ngspice/xyce packages. I'm not actively using any of those but they are generally useful.

## Checklist:

* [x] I want to archive a feedstock:
  * [x] Pinged the team for that feedstock for their input.
  * [x] Supplied a reason for archiving

### DIFF
--- a/requests/archive-simserver.yml
+++ b/requests/archive-simserver.yml
@@ -1,0 +1,5 @@
+action: archive
+feedstocks:
+  - ngspicesimserver-feedstock
+  - jupyterlab-mosaic-feedstock
+  - pyttoresque-feedstock
--- a/requests/archive-simserver.yml
+++ b/requests/archive-simserver.yml
@@ -1,4 +1,6 @@
 action: archive
+reason: >
+  The project has been archived upstream. https://github.com/conda-forge/admin-requests/pull/1879/
 feedstocks:
   - ngspicesimserver-feedstock
   - jupyterlab-mosaic-feedstock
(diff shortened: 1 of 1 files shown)


carterbox:
@conda-forge/pyttoresque @conda-forge/jupyterlab-mosaic @conda-forge/ngspicesimserver 

> This should remove an mosaic/pyttoreque/simserver packages.

We don't remove packages from the channel. Archiving is appropriate when the upstream project is abandoned or discontinued, but it doesn't remove packages. It just disables the feedstock.

Looks like this is only the case for https://github.com/NyanCAD/NgspiceSimServer and https://github.com/NyanCAD/jupyterlab-mosaic

Unless not-archiving https://github.com/NyanCAD/Pyttoresque was an accidental omission?

carterbox (changes requested):
Please also open an issue in each of these feedstocks explaining why it is archived (as mentioned in the checklist). Once the feedstock is archived, it will be frozen, so it is helpful to explain the reason for archive before then.

pepijndevos:
I'm not spending more time on this. They can just linger on forever if you'd prefer.

carterbox:
@pepijndevos Is https://github.com/NyanCAD/Pyttoresque discontinued or not?

pepijndevos:
> @pepijndevos Is https://github.com/NyanCAD/Pyttoresque discontinued or not?

Yes.

Thank you.

PR #1956, opened 2026-03-23
Automatically file archive issue

### LINKED ISSUE #1324: Add logic to archive request to open issue automatically
Currently we ask folks to open an issue on a feedstock before archiving and link it here. Though am wondering if we could simply fold that into the archive process. We could add a field to the request so that users could fill out the issue description. Also we could standardize the issue titles and xref the archive request

Think this would help simplify the archive effort here and improve traceability for future readers

mgorny:
Add a `reason` field that can be used to file the archiving issue on the feedstock automatically.

Fixes #1324

### DIFF
--- a/conda_forge_admin_requests/archive_feedstock.py
+++ b/conda_forge_admin_requests/archive_feedstock.py
@@ -5,7 +5,7 @@ import requests
 from .utils import GH_ORG, get_gh_headers, raise_json_for_status
 
 
-def process_repo(repo, task):
+def process_repo(repo, task, reason=None):
     owner = GH_ORG
     headers = get_gh_headers()
 
@@ -29,6 +29,18 @@ def process_repo(repo, task):
         print(f"feedstock {repo} is already {target_status}", flush=True)
         return
 
+    if task == "archive" and reason is not None:
+        r = requests.post(
+            f"https://api.github.com/repos/{owner}/{repo}/issues",
+            headers=headers,
+            json={
+                "title": "Archive the feedstock",
+                "body": reason,
+            },
+        )
+        raise_json_for_status(r)
+        print(f"archival issue created: {r.json()['html_url']}", flush=True)
+
     r = requests.patch(
         f"https://api.github.com/repos/{owner}/{repo}",
         headers=headers,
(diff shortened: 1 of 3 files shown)


mgorny (commented):
This is the first approximation. I've inlined GitHub API since the script already did that, and the relevant endpoints are trivial to use.

h-vetinari (commented):
What I'd almost consider more important than having a skeleton issue is to have a note in the README that adds language saying "this has been archived, if you need it, open a PR to unarchive on https://github.com/conda-forge/admin-requests"; in an ideal world, the README text would also refer to the admin-requests PR that triggered the archival (where people could then leave references to further feedstock issues if they exist).

h-vetinari:
> As in committing straight to the feedstock repository?

Yes.

> And then removing that line when unarchiving the feedstock, or just assuming the next rerender will take care of that?

I think it'd be fine to let a rerender take care of it. Could be part of the text in the README, à la:

> This feedstock has been archived (see associated [request](https://github.com/conda-forge/admin-requests/pull/1956) and discussion/references therein). If this feedstock needs to be unarchived, please file an appropriate PR to https://github.com/conda-forge/admin-requests (with an explanation why un-archival is desirable). Then open a PR to this feedstock that updates the feedstock as necessary; don't forget to rerender the feedstock (either using the appropriate bot command in the PR, or by hand in your local checkout), which will also remove this blurb.

jaimergp:
What about leaving the README alone and instead editing the repository description in the sidebar pointing to the issue?

h-vetinari:
> What about leaving the README alone and instead editing the repository description in the sidebar pointing to the issue?

Completely fine by me as an approach!

carterbox:
Let's not perfect be the enemy of good here. I believe the remaining unaddressed comments are requests to increase the original scope of this MR. These include modifying the feedstock description. They may be addressed in a future MR.

conda_forge_admin_requests/archive_feedstock.py
```diff
@@ -31,6 +31,18 @@ def process_repo(repo, task):
         print(f"feedstock {repo} is already {target_status}", flush=True)
         return
 
+    if task == "archive" and reason is not None:
+        r = requests.post(
+            f"https://api.github.com/repos/{owner}/{repo}/issues",
+            headers=headers,
+            json={
+                "title": "Archive the feedstock",
+                "body": reason,
```
carterbox:
```suggestion
                "body": f"If you need to unarchive this feedstock, open a PR to unarchive at https://github.com/conda-forge/admin-requests. This feedstock has been archived for the following reason: {reason}",
```

carterbox (approved):

h-vetinari:
Please leave some time for others to respond, especially when reviving old PRs.

PR #1980, opened 2026-04-01
Mark sluarch v2024.5.16 as broken

sjg2203:

## Checklist:

# Reason
Package version was yanked and removed from pypi: https://pypi.org/project/sluarch
close #1975 
ping @conda-forge/sluarch


hmaarrfk:
So what happened here exactly? can you please explain WHY it was removed from pypi, and why the repo no longer exists on github?

sjg2203:
> So what happened here exactly? can you please explain WHY it was removed from pypi, and why the repo no longer exists on github?

This version contains errors and should be removed from conda, I also wanted to adhere to semantic versioning. Therefore, I removed it from pypi, while I work on these toolboxes (cf the other PR I submitted, there are issues that require a reset to the previous version), I would prefer if these were removed.

hmaarrfk:
Can you please cross reference the new repo

sjg2203:
> Can you please cross reference the new repo


Feedstocks and repos remain unchanged

hmaarrfk:
Please be careful next time. Yanking packages like this is really frowned upon at conda-forge

---------------------------------------------

Upstream author is the conda-forge maintainer.
They released 2 small versions, likely for themselves, so I'm going to honor their request to mark their package here as broken.

https://github.com/sjg2203/SluArch

PR #1989, opened 2026-04-07
mark proj 9.8 as broken

ocefpaf:

## Checklist:

Latest proj is broken and a 9.8.1 version will be released soon. For now we should remove this to prevent bad/wrong results when using it. See https://github.com/OSGeo/PROJ/pull/4736 for more info.

Pinging @conda-forge/proj-4

### DIFF
--- a/requests/proj_9.8.yml
+++ b/requests/proj_9.8.yml
@@ -1,0 +1,7 @@
+action: broken
+packages:
+  - win-64/proj-9.8.0-hd30e2cd_0.conda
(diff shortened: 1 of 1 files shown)


xylar (approved):

akrherz:
@ocefpaf Looking at this, the broken packages omitted linux64 ?
Screenshot from 2026-04-08 21-51-31

ocefpaf:
Ack. Do you want to send a PR fixing that? If not, I'll get to it as soon as I get back to my laptop.

PR #1999, opened 2026-04-08
add pyami-core output: pyami-core-cli

kratsg:
## Checklist:

This is to split the `pyami-core` into a python-only output and an output that ships the `cli` separately under `ami`. This will be needed for downstream packages that override the `ami` cli by default that comes from `pyami-core` (this is how `pyami` was designed in the first place).


kratsg:
I am the core team for @conda-forge/pyami-core .

@conda-forge/core, this is ready for review. Let me know if you have any questions. 👍

PR #2000, opened 2026-04-08
add qtkeychain output: qtkeychain-qt*

gillins:
## Checklist:

ping @conda-forge/qtkeychain @SrNetoChan

xref: https://github.com/conda-forge/qtkeychain-feedstock/pull/28

For QGIS we need to support both Qt5 (LTR) and Qt6 (latest) builds for now. This PR adds the ability to build qtkeychain both as Qt5 and Qt6. The Qt6 build will be `qtkeychain-qt6` and the Qt5 build will stay unchanged as `qtkeychain`. I've added a wildcard so we don't have to do this again next time there is a Qt upgrade.


requests/add-qtkeychain-qtx.yml
```diff
@@ -0,0 +1,4 @@
+action: add_feedstock_output
+feedstock_to_output_mapping:
+  qtkeychain:
+    - "qtkeychain-qt*"
```
hmaarrfk:
```suggestion
    - "qtkeychain-qt6"
```

Qt is pretty unlikely to update again. 

I really dislike wildcards due to their unforeseen consequences.

hmaarrfk:
Qt5 is legacy. I removed the wildcard

PR #2007, opened 2026-04-10
Do not register blacksmith or namespace by default

jaimergp:
I keep forgetting about this part, sorry about the noise


jaimergp:
Merged quickly so it applies before the next batch of feedstocks are created after the fix in https://github.com/conda-forge/staged-recipes/commit/500f0c644faa05eeebc41dd058fa1ab02958f940

PR #2013, opened 2026-04-13
mark codex as broken for win-arm64

pavelzw:

## Checklist:

  ping @conda-forge/codex

https://github.com/conda-forge/codex-feedstock/pull/117#discussion_r3071576116

the win-arm64 packages are empty for codex:

```
pixi exec pixi-browse -m codex --platform win-arm64
```

image

```
❯ pixi exec rattler inspect https://conda.anaconda.org/conda-forge/win-arm64/codex-0.118.0-h50586dc_0.conda
name: codex
version: 0.118.0
build: h50586dc_0
license: Apache-2.0
subdir: win-arm64
depends:
  - vc >=14.3,<15
  - vc14_runtime >=14.44.35208
paths: (0 total)
```

interestingly, not all packages are broken, only a couple


zelosleone:
It's true codex doesn't build on win-arm64 correctly, i also tried to build in the past (mainly on codex feedstock with conda-forge servers) and it never worked correctly due to dependencies it requires not being available on win64-arm. (nothing to do with rust alone.)

xhochy (approved):

PR #2024, opened 2026-04-16
MNT: Add unpatched version and ZIG support

MementoRC:
## Checklist:

QEMU-execve applies a unsupported patch to add redirection of `execve()` calls to itself rather than to the native dynamic loader. It may be useful for users to have access to the un-patched QEMU.

The `qemu-zig` adds additional ZIG patches to enhance QEMU for more integrated use when compiling with the ZIG compiler. It is directed to be installed in `${PREFIX}/lib/qemu-zig/` folder that conda ZIG has been upgraded to use

Related PR:
https://github.com/conda-forge/qemu-execve-feedstock/pull/23

### DIFF
--- a/requests/qemu-execve.yaml
+++ b/requests/qemu-execve.yaml
@@ -1,0 +1,11 @@
+action: add_feedstock_output
+feedstock_to_output_mapping:
+  qemu-execve:
+    - qemu-user-linux-aarch64
+    - qemu-user-linux-ppc64le
+    - qemu-user-linux-riscv64
+    - qemu-user-linux-s390x
+    - qemu-zig-linux-aarch64
+    - qemu-zig-linux-ppc64le
+    - qemu-zig-linux-riscv64
+    - qemu-zig-linux-s390x


MementoRC:
@conda-forge-admin Ready for review

MementoRC:
Don't bother

h-vetinari:
> @conda-forge-admin Ready for review

this ping went nowhere (it's a bot that doesn't read notifications)

> Don't bother

If this is just about the response times, please be a bit more patient. On the other side there's nothing more than a few volunteers trying to keep many balls up in the air.

MementoRC:
> 
> If this is just about the response times, please be a bit more patient. On the other side there's nothing more than a few volunteers trying to keep many balls up in the air.

It is not 'just about response times', the main reason to expand the recipe was the need to have a custom zig version. The new qemu v11 (from just a few days ago) incorporates that support. It is, thus, simpler to provide both vanilla and execve-patched under the same recipe - The closure is rational.

PR #2033, opened 2026-04-21
mark libprotobuf builds with wrong major version as broken

h-vetinari:
We published libprotobuf with the wrong version number (c.f. https://github.com/conda-forge/libprotobuf-feedstock/pull/290), mark those builds as broken. They're not widely used because we haven't migrated anything for this yet.


hmaarrfk (approved):

h-vetinari:
thanks!

PR #2034, opened 2026-04-22
feat: reset recipe-maintainers on feedstock archive/unarchive

chrisburr:
Archive requests now clear `extra.recipe-maintainers` (pushed to the default branch before the repo is flipped to archived). Unarchive requests take a `{feedstock: [new-maintainers]}` mapping and replace the maintainers list with the provided usernames after unarchiving; the old shape and empty maintainer lists are rejected at check time, and each username is validated against the GitHub users API.

### DIFF
--- a/README.md
+++ b/README.md
@@ -48,7 +48,15 @@ If you want to request a feedstock to be archived, send a Pull Request
 adding a new `.yml` file in `requests` folder with a list of the feedstock names
 without `-feedstock`. See `examples/example-archive.yml` for an example.
 (e.g., for `python-feedstock`, the feedstocks list must contain `python`).
-For unarchiving, see `examples/example-unarchive.yml` for an example.
+For unarchiving, see `examples/example-unarchive.yml` for an example. Unarchive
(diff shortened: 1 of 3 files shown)


h-vetinari:
What's the concern about removing maintainers before archiving? They can't do anything anyway once the feedstock is archived...? I could understand a check about whether maintainers are still active before unarchiving, but I don't much see the point to do this before archival...

In #1956, we had discussed pushing commits (for README updates etc.) before archival, and @jaimergp had preferred a solution that didn't involve pushes. I could live with them, but they should provide a tangible "return on investment" in some sense. If we do end up pushing changes for whatever reason, I believe we should combine those commits into one à la: `[ci skip] Archive foo-feedstock` and then do all the necessary things as part of that (and have a single HEAD commit to revert when unarchiving[^1])

[^1]: modulo logic that updates maintainers, etc.

beckermr:
The difference here is permissions, but I think it is fine to push to the readme as well. So let's do both.

PR #2044, opened 2026-04-24
Document storage offered in each provider

jaimergp:
Sources:

- https://namespace.so/docs/architecture/compute/machine-shapes#storage-allocation
- https://docs.blacksmith.sh/blacksmith-runners/overview#runner-tags


beckermr:
Oh woops I merged another PR by @jschueller. 🤦

PR #2048, opened 2026-04-25
Add spfft-static output to spfft

mkrack:
Add spfft-static output name for static `spfft` library build in addition to the existing shared library build `spla`.

## Checklist:

### DIFF
--- a/requests/add_spfft-static_output.yaml
+++ b/requests/add_spfft-static_output.yaml
@@ -1,0 +1,4 @@
+action: add_feedstock_output
+feedstock_to_output_mapping:
+  spfft:
+    - spfft-static


hmaarrfk:
Why use static libraries?

mkrack:
> Why use static libraries?

Good question! I have to confess that I have no strong use case currently. SpFFT (like SpLA) is used by [SIRIUS](https://github.com/electronic-structure/SIRIUS) and [CP2K](https://github.com/cp2k/cp2k) as low-level library mostly in HPC applications. I thought that it could be useful to have also statically linked stand-alone versions of SpFFT (and SpLA) for testing and benchmarking in HPC.

Thanks for merging the request!

hmaarrfk:
Ok. It might be useful to have stsatic libraries outside the conda ecosystem. 

But within it I’ve found it is of limited value. 

Thanks for explaining and helping the coda ecosystem.

PR #2077, opened 2026-05-12
Add token reset action for enfrosp feedstock

danschef:
For the enfrosp feedstock, two new builds were successfully created in https://github.com/conda-forge/enfrosp-feedstock/pull/9 and https://github.com/conda-forge/enfrosp-feedstock/pull/10. However, they don´t show up at https://anaconda.org/channels/conda-forge/packages/enfrosp/files, even after re-rendering the feedstock and waiting for days. So maybe a token reset helps. If not, I would appreciate any advice.


xhochy:
The CI on main time outs, so a token reset won't help.

danschef:
Thanks @xhochy, I missed that. Will see how to fix that.

PR #2098, opened 2026-05-19
FEA: zig supporting libs

MementoRC:
## Checklist:

@wolfv as agreed

This adds limited scoped supporting libraries for zig decoupling from GCC/LLVM(stdc++)
zig-libxml2, zig-zstd, zig-zlib are intended only for RISCV64 until arch builds are available on conda-forge

### DIFF
--- a/requests/zig.yaml
+++ b/requests/zig.yaml
@@ -5,7 +5,3 @@ feedstock_to_output_mapping:
     - zig-llvm
     - "zig-libllvm*"
     - zig-tblgen
-    # Temporarily for riscv64
-    - zig-libxml2
-    - zig-zlib
-    - zig-zstd
--- a/requests/zig.yaml
+++ b/requests/zig.yaml
@@ -1,0 +1,11 @@
+action: add_feedstock_output
+feedstock_to_output_mapping:
+  zig:
+    - zig-libcxx
+    - zig-llvm
+    - "zig-libllvm*"
+    - zig-tblgen
+    # Temporarily for riscv64
+    - zig-libxml2
+    - zig-zlib
+    - zig-zstd


isuruf (changes requested):
This is basically vendoring other packages and we don't allow it in conda-forge unless there's a good reason. Packaging `zig-libxml2` just because riscv64 does not have it yet is not how we do things at conda-forge. If support for an upstream conda package is not there, we don't vendor it: we fix it in the upstream conda package.

MementoRC:
> we fix it in the upstream conda package

Would you accept a PR adding libxml2, zlib, zstd builds for riscv64 in their respective feedstocks?
Would you accept a PR on the llvm-feedstock to add a zig-llvm, zig-libcxx output?

isuruf:
> Would you accept a PR adding libxml2, zlib, zstd builds for riscv64 in their respective feedstocks?

Yes, when we have the compilers and the migrator ready.

>  Would you accept a PR on the llvm-feedstock to add a zig-llvm, zig-libcxx output?

I don't know the difference between llvm and zig-llvm. You would have to open an issue.

MementoRC:
> difference between llvm and zig-llvm

ZIG C++ library is `libc++` and in some cases conflicts arise due to conda-forge LLVM using GCC/stdc++ (for example, zig cannot compile itself completely). A solution to this is a 'lightweight LLVM' with the appropriate targets for conda-forge, compiled fully shared with zig and ideally located in a custom path

isuruf:
> for example, zig cannot compile itself completely

What are the advantages of doing this?
Is this the only use case of having a LLVM compiled with libcxx? conda-forge cannot support building each and every package with libc++ and we must stay with libstdc++.

MementoRC:
> What are the advantages of doing this?

zig would be decoupled from GCC/Clang/stdc++

> each and every package with libc++

I am not understanding why conda-forge would have to do that? User preferring ZIG would have all dynamic libraries available with the `zig-` prefix.
For example, we could create a cross-compiled zig for win-arm64 (currently it fails due to `clang 20.*` libs not available)

isuruf:
@MementoRC, we have to think about the whole ecosystem in general. One package going in its own way is not really sustainable. I want to have some discussion about your needs for zig, but you have to help out with other feedstocks to make it work properly instead of hacking your way through.

PR #2099, opened 2026-05-19
Add output runner-intel to feedstock conda-forge/runner-feedstock

lxknll:
## Checklist:

ping @conda-forge/runner-feedstock

**Description:**
Adding `runner-intel` as a new authorized output to the `runner-feedstock`. This allows us to publish a parallel, Intel-optimized variant compiled with `ifx` and `mkl` alongside our standard GNU/OpenBLAS build, all from the same feedstock.

### DIFF
--- a/requests/runner_add_intel.yml
+++ b/requests/runner_add_intel.yml
@@ -1,0 +1,4 @@
+action: add_feedstock_output
+feedstock_to_output_mapping:
+  runner:
+    - runner-intel


xhochy:
You would normally do this with a change in the build string (e.g. like we are differentiating the CPU and GPU packages in tensorflow or jaxlib) instead of a second output.

lxknll:
Thank you for the feedback @xhochy. I did not know that this is the proper way to do it and will give it a try. Once I confirm that it is working as intended, I will close this PR.

lxknll:
Thanks again for the hint to use multiple variants - this works very nicely. Closing this PR.

PR #2110, opened 2026-05-29
Extend scikit-learn-intelex feedstock for gpu package

david-cortes-intel:
Adds entry `scikit-learn-intelex-gpu` to this feedstock:
https://github.com/conda-forge/scikit-learn-intelex-feedstock

This is because the library is being split into different packages for CPU and GPU, which will both be produced from the same source code and recipe.

## Checklist:

@conda-forge/scikit-learn-intelex

### DIFF
--- a/requests/add-scikit-learn-intelex-output-scikit-learn-intelex-gpu.yaml
+++ b/requests/add-scikit-learn-intelex-output-scikit-learn-intelex-gpu.yaml
@@ -1,0 +1,4 @@
+action: add_feedstock_output
+feedstock_to_output_mapping:
+  scikit-learn-intelex:
+    - scikit-learn-intelex-gpu


xhochy:
1. Please always reference a PR here
2. Adding another name is not the right way to approach this as the builds will probably conflict. Instead, make a different build string like we do in pytorch, tensorflow, jax.

david-cortes-intel:
Thanks for the follow up.

We are still working on crafting a new recipe, but this would be the sketch (currently in an internal recipe from the source repository):
https://github.com/uxlfoundation/scikit-learn-intelex/pull/3165

> Adding another name is not the right way to approach this as the builds will probably conflict. Instead, make a different build string like we do in pytorch, tensorflow, jax.

The approach we'd like to follow is to have a GPU package that would contain only GPU-related files and depend on the CPU package to run, while the CPU package would be standalone. We'd like to follow this approach because that's how it's being done for the PyPI packages where build variants are not supported:
https://pypi.org/project/scikit-learn-intelex-gpu

Do you think we could use this same approach in conda-forge?

xhochy:
This sounds OK (not perfect but I don't have a better suggestion). I would only merge here, though once the PR to the feedstock is up.

ethanglaser:
This is a follow-up to https://github.com/conda-forge/admin-requests/pull/2078, which was set up similarly

ethanglaser:
The feedstock PR is drafted, I believe it is pending this merge: https://github.com/conda-forge/scikit-learn-intelex-feedstock/pull/80

ethanglaser:
@xhochy please let us know if anything else is necessary to finalize this

xhochy (approved):

PR #2111, opened 2026-05-29
Mark PySCF broken

BastianZim:

## Checklist:

I am one of the maintainers and we decided this in https://github.com/conda-forge/pyscf-feedstock/pull/61

For some unfortunate reason, PR [#60](https://github.com/conda-forge/pyscf-feedstock/pull/60) caused the variable microarch_level to be a string rather than an int, causing comparisons like `- if: unix and x86_64 and (microarch_level < 4)` to fail. This is now fixed, and we have verified that the correct -march flags are passed to the compiler.


chillenb:
I've double checked the list of packages and it looks good to me.
The most important ones are all pyscf=2.13.0 packages with a build number ending in 2, and these are all included.

chillenb:
That was extremely fast!!

PR #2112, opened 2026-05-30
Request python bindings outputs added to roboplan-feedstock

sea-bass:
These new outputs are being added to `roboplan-feedstock` because the way that Python bindings are generated is switching from one monolithic output (`roboplan-python`) to per-package Python binding outputs.

So for example, the `roboplan-simple-ik` package has both a C++ only library `libroboplan-simple-ik` and a Python bindings output `roboplan-simple-ik-python`.

## Checklist:

### DIFF
--- a/requests/roboplan.yaml
+++ b/requests/roboplan.yaml
@@ -2,7 +2,6 @@ action: add_feedstock_output
 feedstock_to_output_mapping:
   roboplan:
     - roboplan-example-models-python
-    - roboplan-python
     - roboplan-oink-python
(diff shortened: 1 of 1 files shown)


requests/roboplan.yaml
```diff
@@ -0,0 +1,9 @@
+action: add_feedstock_output
+feedstock_to_output_mapping:
+  roboplan:
+    - roboplan-example-models-python
+    - roboplan-python
```
sea-bass:
```suggestion
```

PR #2136, opened 2026-06-08
Add pytrilinos to trilinos outputs

rfiorella:
Instead of essentially building all of the trilinos pacakges twice, suggest archiving the current pytrilinos feedstock and adding a pytrilinos* output to the trilinos feedstock. Glob form of pytrilinos is used to help distinguish between pytrilinos and pytrilinos2, the latter of which is the more modern package.

ping @conda-forge/trilinos 

## Checklist:

### DIFF
--- a/requests/add-feedstock-output.yml
+++ b/requests/add-feedstock-output.yml
@@ -1,0 +1,5 @@
+action: add_feedstock_output
+feedstock_to_output_mapping:
+  trilinos:
+    - trilinos
+    - "pytrilinos*"
--- a/requests/add-feedstock-output.yml
+++ b/requests/add-feedstock-output.yml
@@ -1,4 +1,4 @@
 action: add_feedstock_output
 feedstock_to_output_mapping:
   trilinos:
-    - "pytrilinos*"
+    - pytrilinos
--- a/requests/add-feedstock-output.yml
+++ b/requests/add-feedstock-output.yml
@@ -1,5 +1,4 @@
 action: add_feedstock_output
 feedstock_to_output_mapping:
   trilinos:
-    - trilinos
     - "pytrilinos*"


requests/add-feedstock-output.yml
```diff
@@ -0,0 +1,5 @@
+action: add_feedstock_output
+feedstock_to_output_mapping:
+  trilinos:
+    - trilinos
+    - "pytrilinos*"
```
h-vetinari: let's avoid globs but explicitly name packages. Also, I'm not in favour of a `pytrilinos2`. Let's use other techniques for `pytrilinos` users to transition smoothly

requests/add-feedstock-output.yml
```diff
@@ -0,0 +1,5 @@
+action: add_feedstock_output
+feedstock_to_output_mapping:
+  trilinos:
+    - trilinos
```
h-vetinari: no need to add the packages that the feedstock is already allowed to publish

requests/add-feedstock-output.yml
rfiorella: I can change this, but I realized I was a bit ambiguous in my prior messages: to clarify, I am not suggesting `pytrilinos2` as a way to distinguish from the now archived `pytrilinos` output from the `pytrilinos-feedstock`; rather, the name of the package that provides the more modern python bindings in trilinos is named `pytrilinos2`.

I could keep the output with the same name though and just update the cmake to build the new bindings in future updates to `trilinos-feedstock` if it's preferable to keep the same package name here

h-vetinari: We don't _have_ to rename the package, but we still _can_, I see that upstream is [indeed](https://github.com/search?q=repo%3Atrilinos%2FTrilinos+%2Ffrom+PyTrilinos%2F&type=code) using `PyTrilinos2`.

Normally I'd suggest to go with the naming on PyPI, but I cannot even find that package on PyPI? So unless there's a strong reason, my suggest would be to focus on `pytrilinos` for now. This solves various things that would otherwise be a pain to handle, like avoiding co-installability with (pre-v2) `pytrilinos` artefacts.

If you really want to enable `pytrilinos2`, we can add it here as an allowed output for now, but not with a glob.

rfiorella: Let's leave it as just pytrilinos for now and we can decide on pytrilinos2 later as I work to get the built version of trilinos up to more current releases

h-vetinari (approved):

PR #2141, opened 2026-06-10
mark all but version 1.1.1 as broken b/c those do not come from tags

ocefpaf:
Marking thesepackages as broken b/c only version 1.1.1 has an updstream tag and correct numpy pinning.


ocefpaf:
pre-commit.ci autofix

PR #2144, opened 2026-06-14
allow `libharfbuzz{,-devel}` for harfbuzz

h-vetinari:
For https://github.com/conda-forge/harfbuzz-feedstock/pull/162


hmaarrfk:
I’m not sure what we should do about the global pin. Maybe just remove it altogether since harfbuzz never breaks abi

h-vetinari:
> I’m not sure what we should do about the global pin. Maybe just remove it altogether since harfbuzz never breaks abi

For now it doesn't hurt to keep it, IMO

PR #2164, opened 2026-06-24
Add ambertools to depot

mikemhenry:
## Checklist:

### DIFF
--- a/requests/ambertools-depot.yml
+++ b/requests/ambertools-depot.yml
@@ -1,0 +1,4 @@
+action: depot
+feedstocks:
+  - ambertools
+revoke: false


mikemhenry:
@conda-forge/ambertools this is for https://github.com/conda-forge/ambertools-feedstock/pull/181

I've got aarch64 builds working locally using the feedstock (builds uploaded here https://anaconda.org/channels/mmh/packages/ambertools/files) but the aarch64 builds are timing out, see https://github.com/conda-forge/ambertools-feedstock/pull/181#issuecomment-3848679936

Thanks!

h-vetinari (approved):

jaimergp:
@mikemhenry we have native ARM runners for Linux now, no need for extra resources I'd say?

mikemhenry:
Yes, it looks like I won't need extra resources. I'd like to keep them enable if I need them but after I get ambertools 25 & 26 built I can open a PR to revoke access to the extra resources if we don't need them.

PR #2172, opened 2026-06-28
Add graph-tool to namespace

count0:
## Checklist:

This is a change of a previously accepted request from blacksmith to namespace, since blacksmith appears to be failing for some unknown reason.

ping @conda-forge/graph-tool

### DIFF
--- a/requests/graph-tool.yml
+++ b/requests/graph-tool.yml
@@ -1,4 +1,4 @@
-action: blacksmith
+action: namespace
 feedstocks:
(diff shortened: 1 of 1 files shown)


h-vetinari (approved):
Let's try if this works better than blacksmith...

count0:
> Let's try if this works better than blacksmith...

It still seems to be 'failing'... Is it possible to get any insight on the nature of the failure?

PR #2200, opened 2026-07-14
Do not push empty commits in main requests workflow

### LINKED ISSUE #2117: Empty commits from failing token resets can pile up enormously
I'm just noticing lots of failed token reset commits here. CC @beckermr @jaimergp

jaimergp:
- Closes https://github.com/conda-forge/admin-requests/issues/2117. 
- Two parts to this:
	- First commit stops the chain of empty commits. The cronjob will retry anyway in the next cycle, so no need to re-push with the same contents.
	- 2nd and following commits implement the error reporting. If the failing requests have been there for long enough (>6h), an issue is created for each one if none exists yet.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -82,3 +82,19 @@ jobs:
         run: |
           git pull
           git push
+
+      - name: raise issue if needed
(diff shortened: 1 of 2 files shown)


beckermr (approved):

jaimergp:
Let's give this a try. I'll introduce a purposefully broken request in a subsequent PR to see how it works in `main`.

PR #2201, opened 2026-07-14
Check request before running, return deepcopied update with failures

jaimergp:
I noticed that some actions are not checking before running, or not returning the updated requests with the failures. In the case of `archive`, it was always returning None, which is considered "success".

2nd commit is just a cleanup.

### DIFF
--- a/conda_forge_admin_requests/archive_feedstock.py
+++ b/conda_forge_admin_requests/archive_feedstock.py
@@ -1,4 +1,6 @@
-import subprocess
+from __future__ import annotations
+
+import copy
 
 import requests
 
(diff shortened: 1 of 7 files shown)


conda_forge_admin_requests/archive_feedstock.py
```diff
@@ -52,9 +55,11 @@ def run(request):
             pkgs_to_do_again.append(feedstock)
 
     if pkgs_to_do_again:
-        request["feedstocks"] = pkgs_to_do_again
-
-    subprocess.check_call(["git", "show"])
+        request = copy.deepcopy(request)
+        request["packages"] = pkgs_to_do_again
```
beckermr: Why did the dict key change here?

beckermr (changes requested):

conda_forge_admin_requests/archive_feedstock.py
jaimergp: Oops, excellent catch, thanks!
```suggestion
        request["feedstocks"] = pkgs_to_do_again
```

beckermr (approved):

PR #2209, opened 2026-07-17
NEW: Request GHA runners for numba-cuda-mlir

carterbox:
## Checklist:

@conda-forge/cuda-python @conda-forge/numba-cuda-mlir 

We want to build on native ARM runners because maintaining a cross-compiling build-script for LLVM components is complicated.

### DIFF
--- a/requests/numba-cuda-mlir.yml
+++ b/requests/numba-cuda-mlir.yml
@@ -1,0 +1,4 @@
+action: depot
+feedstocks:
+  - numba-cuda-mlir
+revoke: false


chrisburr:
Native arm runners are already available to all feedstocks for Linux and macOS.

xhochy:
This will suffice:

```
provider:
  linux_aarch64: default
  osx_arm64: default
```

carterbox:
That's great news! I figured it wasn't the case because the docs still state that emulation is the default. https://conda-forge.org/docs/how-to/advanced/enable-archs/#manual-migrations

PR #2219, opened 2026-07-22
Mark langflow/lfx/langflow-sdk 1.10.3 as broken (built from unreleased tag)

pb01ka:
This request marks the langflow 1.10.3 artifacts (and the related `lfx` 1.10.3 and `langflow-sdk` 0.2.3 outputs) as broken on the `conda-forge` channel.

The langflow-feedstock autotick bot proposed `v1.10.3` (and I merged by trusting it), which is an upstream **git tag that was never cut as a GitHub Release nor published to PyPI**:

- v1.10.3 is **not** a GitHub Release
- The version is **absent from PyPI**

The newest genuine upstream release is `v1.10.2`, so these 1.10.3 artifacts should never have been shipped.

> **Note:** `langflow-sdk-0.2.2` is intentionally **not** included. It was built from the
> released v1.10.2 source and is legitimate.

### Why mark broken instead of patching the repo data

First of all, I would like to state that the README rightly prefers repo-data patches over marking broken, since patching keeps packages installable and environments reproducible.

However, I think this case is different: 1.10.3 is an **entire version that should never have existed**, built from an unreleased upstream tag - there is no metadata to correct, so a patch has nothing to fix. I think, marking broken is the appropriate action here, and it does not delete the artifacts; only new solves stop picking up a version upstream never shipped.

### Preventing recurrence

In https://github.com/conda-forge/langflow-feedstock/pull/7, the feedstock's version source has been pinned to `githubreleases` and the recipe reverted to 1.10.2, so the bot will no longer pick up unreleased git tags.

### Impact

These are low-download packages used mainly by the maintainers, consistent with the "small number of users" (~100 - I hope its a small enough number) guideline in the admin-requests README for marking artifacts broken.

---

### Guidelines for marking packages as broken:

## Checklist:

  * [x] Pinged the team for the package for their input - I am one of the maintainers of this package. My co-maintainer also thinks the same - https://github.com/conda-forge/langflow-feedstock/pull/7#issuecomment-5033537856

### DIFF
--- a/requests/broken-langflow-1.10.3.yml
+++ b/requests/broken-langflow-1.10.3.yml
@@ -1,0 +1,9 @@
+action: broken
+packages:
+- noarch/langflow-1.10.3-pyhc5c1524_0.conda
+- noarch/langflow-1.10.3-pyhaf0af83_1.conda
+- noarch/langflow-base-1.10.3-pyh7de08d0_0.conda
+- noarch/langflow-base-1.10.3-pyhf1e8ddc_1.conda
+- noarch/lfx-1.10.3-pyh3edcccd_0.conda
+- noarch/lfx-1.10.3-pyh62f3024_1.conda
+- noarch/langflow-sdk-0.2.3-pyhc364b38_1.conda


pb01ka:
@conda-forge/core Waiting for your reviews. TY.

xhochy (approved):

PR #2223, opened 2026-07-24
Mark ipyrad2 as broken

isaacovercast:
I am requesting to mark all ipyrad2 assets in conda-forge as broken because its required runtime dependencies are available only from Bioconda. Although installation from conda-forge succeeds, the resulting environment is not functional. I would like to move the package to Bioconda. Please move this artifact to the broken label to prevent new installations while preserving reproducibility.

Following a workflow suggested to me on the [conda-forge zulip ](https://conda-forge.zulipchat.com/#narrow/channel/457337-general/topic/.E2.9C.94.20Package.20needs.20to.20move.20from.20conda-forge.20to.20bioconda/with/611352024), I have requested the [conda-forge/ipyrad2-feedstock](https://github.com/conda-forge/ipyrad2-feedstock) be archived, and it was archived today. Now I am requesting ipyrad2 artifacts be marked as broken, so that I can publish to bioconda. I am the package maintainer, and this is a new package so there are no current users of this package except me.

## Checklist:

### DIFF
--- a/requests/ipyrad2.yml
+++ b/requests/ipyrad2.yml
@@ -1,0 +1,3 @@
+action: broken
+packages:
+- noarch/ipyrad2-0.1.9-pyhd8ed1ab_0.conda


isaacovercast:
@conda-forge/core Anything I can do/need to do to get this marked as broken so I can migrate to bioconda? Thanks so much

isaacovercast:
Can someone please help me understand if there is something I haven't done correctly to get this package marked as broken? I feel like I have followed the process, but maybe I missed an important detail. The PR idling for one month makes me feel like I haven't checked a needed box. Any clues? Thanks so much.

h-vetinari:
> Can someone please help me understand if there is something I haven't done correctly to get this package marked as broken?

You did things correctly, apologies this fell through the cracks.

PR #2256, opened 2026-08-11
Groundingdino py

phreed:
## Checklist:

I have been in conversation with the original feedstock team.
I am now the primary maintainer.

### DIFF
--- a/requests/add-gpu-pkgs.yml
+++ b/requests/add-gpu-pkgs.yml
@@ -3,5 +3,3 @@ feedstock_to_output_mapping:
   groundingdino-py:
     - groundingdino-py
     - groundingdino-py-cuda
-    - groundingdino-py-rocm
-    - groundingdino-py-vulkan
--- a/requests/add-gpu-pkgs.yml
+++ b/requests/add-gpu-pkgs.yml
@@ -1,0 +1,7 @@
+action: add_feedstock_output
+feedstock_to_output_mapping:
+  groundingdino-py:
+    - groundingdino-py
+    - groundingdino-py-cuda
+    - groundingdino-py-rocm
+    - groundingdino-py-vulkan


xhochy:
Can you link a PR please where you add them?

phreed:
I was not sure if I should request the specific package names (which is what I did) or should opt for the generic `groundingdino-py-*`, I opted for the former as I do not expect more than this.

phreed:
> Can you link a PR please where you add them?

https://github.com/conda-forge/groundingdino-py-feedstock/pull/3

I have not added the *-rocm or the *-vulkan yet as that will be a bit of work and require a separate PR.

phreed:
@xhochy Is there a problem with the request?

requests/add-gpu-pkgs.yml
```diff
@@ -0,0 +1,7 @@
+action: add_feedstock_output
+feedstock_to_output_mapping:
+  groundingdino-py:
+    - groundingdino-py
+    - groundingdino-py-cuda
+    - groundingdino-py-rocm
+    - groundingdino-py-vulkan
```
jaimergp: These two are not being used yet, are they?

phreed: That is correct, I am working on the -rocm version, the -vulkan might not be possible.

jaimergp: Let's go one by one then, as they are needed.
```suggestion
```

phreed: I have removed rocm and vulkan until they have been realized.

requests/add-gpu-pkgs.yml
```diff
@@ -0,0 +1,5 @@
+action: add_feedstock_output
+feedstock_to_output_mapping:
+  groundingdino-py:
+    - groundingdino-py
```
jaimergp, line 4: Noting here that this already exists, so maybe we need to remove it from the request, but let's check what happens.

jaimergp (approved):

requests/add-gpu-pkgs.yml
jaimergp:
```
    output groundingdino-py already exists for feedstock conda-forge/groundingdino-py-feedstock
    output groundingdino-py-cuda added for feedstock conda-forge/groundingdino-py-feedstock
```

worked!

PR #2272, opened 2026-08-17
Add albumentationsx subpackages mapping

diegoferigo:
## Checklist:

  * [x] Pinged the relevant feedstock team(s) @conda-forge/albumentationsx 

Enable all `albumentationsx-*` subpackages that will match PyPI's extras (and awaiting the CEP to propagate for a cleaner implementation).

### DIFF
--- a/requests/add_albumentationsx_subpackages.yml
+++ b/requests/add_albumentationsx_subpackages.yml
@@ -1,0 +1,3 @@
+action: add_feedstock_output
+feedstock_to_output_mapping:
+  albumentationsx: "albumentationsx-*"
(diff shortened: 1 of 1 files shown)


requests/add_albumentationsx_subpackages.yml
```diff
@@ -0,0 +1,4 @@
+action: add_feedstock_output
+feedstock_to_output_mapping:
+  albumentationsx:
+    - "albumentationsx-*"
```
jaimergp, line 4:
```suggestion
    - "albumentationsx-with-*"
```

Assuming this is for `extras`, then we usually add this infix. Can you also link to the PR where this is being implemented? Thanks, @diegoferigo!

PR #2273, opened 2026-08-17
Add newton subpackages to feedstock output mapping

diegoferigo:
## Checklist:

  * [x] Pinged the relevant feedstock team(s) @conda-forge/newton 

Enable all `newton-*` subpackages that will match PyPI's extras (and awaiting the CEP to propagate for a cleaner implementation).

### DIFF
--- a/requests/add_newton_subpackages.yml
+++ b/requests/add_newton_subpackages.yml
@@ -1,0 +1,4 @@
+action: add_feedstock_output
+feedstock_to_output_mapping:
+  newton:
+    - "newton-*"


requests/add_newton_subpackages.yml
```diff
@@ -0,0 +1,4 @@
+action: add_feedstock_output
+feedstock_to_output_mapping:
+  newton:
+    - "newton-*"
```
jaimergp, line 4:
```suggestion
    - "newton-with-*"
```

Same here. And add link to PR.

PR #1925, opened 2026-03-08
Add output configuration for libparallelproj feedstock

gschramm:
## Checklist:

* [x] I want to add a package output to a feedstock:
  * [x] Pinged the relevant feedstock team(s)
  * [x] Added a small description of why the output is being added.

PR #2287, opened 2026-08-19
add libpetsird output

casperdcl:
  * [x] Added a small description of why the output is being added

Would like `petsird-feedstock` to produce:

- `noarch/petsird-*py*.conda` (currently works without this PR)
- `osx-arm64/libpetsird-*.conda`
- `linux-64/libpetsird-*.conda`

and in future also:

- `win-64/libpetsird-*.conda`

vis. https://github.com/conda-forge/petsird-feedstock/pull/16
/CC @conda-forge/petsird

### DIFF
--- a/requests/add-libpetsird.yml
+++ b/requests/add-libpetsird.yml
@@ -1,5 +1,4 @@
 action: add_feedstock_output
 feedstock_to_output_mapping:
   petsird:
-    - libpetsird_linux-64
-    - libpetsird_osx-arm64
+    - libpetsird
--- a/requests/add-libpetsird.yml
+++ b/requests/add-libpetsird.yml
@@ -1,0 +1,5 @@
+action: add_feedstock_output
+feedstock_to_output_mapping:
+  petsird:
+    - libpetsird_linux-64
+    - libpetsird_osx-arm64


wolfv:
Where does the `_linux-64` come from? Is this some kind of compiler package?

KrisThielemans:
no, it is a library to read/write some data. @casperdcl knows a lot mroe about conda-forge than me, so he can elaborate why these names were given, but if you could recommend what we should do, it might help. thanks!

casperdcl:
Thanks for checking @wolfv - I'm unsure about the naming; couldn't find any docs. The example given in this repo is:

https://github.com/conda-forge/admin-requests/blob/724f1c436e440284d3750cb4c5d6729fa1826746/examples/example-add-feedstock-output.yml#L3-L10

Presumably most "new output requests" will be about "add `libmypackage` split output to `mypackage-feedstock`", so I interpret the example above as a naming convention `(lib)?mypackage(_targetplatform)?`. Really all unclear to me.

For clarity, we'd like:

`{name}-feedstock` to produce:

- `noarch/{name}-*py*.conda` (currently works without this PR)
- `osx-arm64/lib{name}-*.conda`
- `linux-64/lib{name}-*.conda`

and in future also:

- `win-64/lib{name}-*.conda`

requests/add-libpetsird.yml
```diff
@@ -0,0 +1,5 @@
+action: add_feedstock_output
+feedstock_to_output_mapping:
+  petsird:
+    - libpetsird_linux-64
+    - libpetsird_osx-arm64
```
casperdcl: assuming `target_platform` isn't required:

```suggestion
    - libpetsird
```

Also, if pre-existing working outputs also need to be listed:

```suggestion
    - petsird
    - libpetsird
```

Or based on #1925, this might be accepted despite being against the example's recommendation?

```suggestion
    - libpetsird*
```

Would appreciate confirmation about this @wolfv :)

jaimergp: Most likely you only need `libpetsird` being added (no platform suffix); aka the first suggestion.

PR #2291, opened 2026-08-20
Mark srsly=3.0.0 as broken (yanked upstream)

mgorny:

* You can use `pixi run find-filenames {matchspec}` to get a list of filenames matching given spec.

## Checklist:

Release yanked upstream.

xref https://github.com/conda-forge/srsly-feedstock/issues/60

ping @conda-forge/srsly

### DIFF
--- a/requests/srsly-yanked.yml
+++ b/requests/srsly-yanked.yml
@@ -1,0 +1,31 @@
+action: broken
+packages:
+  - linux-64/srsly-3.0.0-py310hea6c23e_0.conda
+  - linux-64/srsly-3.0.0-py313h7033f15_0.conda
+  - linux-64/srsly-3.0.0-py311h1ddb823_0.conda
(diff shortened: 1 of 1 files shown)


mgorny:
That said, I'm not 100% sure if this is the right thing to do. On one hand, it seems wrong to be pushing a version that was yanked upstream. On the other, the alternative is pretty much patching 2.5.3 to unvendor all libraries and fix tests, which brings us equal to 3.0.0 upstream.

jaimergp:
Ping @conda-forge/spacy

hmaarrfk:
indeed: https://pypi.org/project/srsly/3.0.0/

carterbox:
I believe that the maintainers of the feedstock have decided to revert to 2.x, so marking this package as broken is what needs to occur next.

https://github.com/conda-forge/srsly-feedstock/pull/61

PR #2304, opened 2026-08-27
FEA: Add ocaml support for s390x

MementoRC:
## Checklist:

@conda-forge/ocaml 
The Ocaml implementation and x86_64 cross-compiler to s390x were missing
PR: https://github.com/conda-forge/ocaml-feedstock/pull/102

### DIFF
--- a/requests/ocaml.yaml
+++ b/requests/ocaml.yaml
@@ -1,0 +1,4 @@
+action: add_feedstock_output
+feedstock_to_output_mapping:
+  ocaml:
+    - ocaml_linux-s390x


h-vetinari:
I would prefer to completely remove s390x from our infrastructure and compiler packages. It's an architecture that never achieved enough momentum to get built out, and all the remaining bits we're carrying around are dead weight IMO. 

Cc @conda-forge/ctng-compilers

MementoRC:
@h-vetinari Is this final? 'prefer' and 'CC' suggest an internal decision round

h-vetinari:
Nothing is final, I've just stated my preference on a topic that's been on my mind. Let's wait for feedback from @isuruf, if he wants to keep maintaining the s390x compilers (or other plans for that target).

MementoRC:
It could become of consideration that there never was a conda-forge `ocaml` on s390x. Future parties could find that it is sufficient to at least have one version available. The s390x `ocaml` already pins the last `gcc` v14 for s390x, limiting its future maintenance requirements

MementoRC:
Thank you

PR #2314, opened 2026-09-01
Register libpuma and pumapy outputs for the puma feedstock

EricAtORS:
The puma feedstock is being migrated to a rattler-build recipe that splits the build into libpuma, pumapy and puma. The puma output is already registered, so only the two new outputs need to be added.

Please see https://github.com/conda-forge/puma-feedstock/issues/30

## Checklist:

ping @conda-forge/puma-feedstock 
ping @Tobias-Fischer

### DIFF
--- a/requests/puma-outputs.yml
+++ b/requests/puma-outputs.yml
@@ -1,0 +1,5 @@
+action: add_feedstock_output
+feedstock_to_output_mapping:
+  puma:
+    - libpuma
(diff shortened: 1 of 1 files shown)


carterbox:
Added -devel and -python outputs to your request in case you want to split runtime from development artifacts or want another option for differentiating the python and C APIs.

PR #2317, opened 2026-09-02
Add new packages names for firecrown

marcpaterno:
## Checklist:

### DIFF
--- a/requests/firecrown-add-feedstock-output.yml
+++ b/requests/firecrown-add-feedstock-output.yml
@@ -1,0 +1,8 @@
+action: add_feedstock_output
+feedstock_to_output_mapping:
+  # this entry adds a package name
+  firecrown:
+    - firecrown-deps
+    - firecrown-deps-validated
+    - firecrown-devenv
+    - firecrown-validated


hmaarrfk:
I strongly recommend against this.

what is the point? helping you create a development environment faster?????

conda-forge package releases cycle is too slow for "making it easy to build your development  environment".

marcpaterno:
The problem we are trying to solve is probably not typical for conda packages.

It is important to note that most of the "users" of firecrown are also contributors, and so work with the development environment and not the conda-installed package. Large-scale "production" use will make uyse of the firecrown package, but the vast majority of user interaction is actually with the developer installation, including a clone of the Github repository for firecrown. The main goal for these new packages is helping to provide a robust environment for domain experts who are not experts in package management. We want to make it easy for such users to add new packages to their (development) environment without disrupting the versions of packages already installed.
This is important for reproducibility of scientific results.

An additional critical goal is providing flexibly for those who have the expertise and need to update the versions of already-installed packages. Even though this loses some guide-rails for reproducibility, it is sometimes needed for adding new functionality.

So the importance not primarily "creating a development environment faster" (the conda-lock files are already pretty good at that) but rather "making the development environment robust for domain experts who are not experts in package management".

For development, we currently advise users to create a conda environment from environment.yml. However, once the environment is created, the dependency constraints from that file are no longer enforced. If the user later installs or updates another package, conda may change dependencies in ways that move the environment away from the versions we test in CI.

The purpose of firecrown-validated would be to provide an installable environment specification whose dependency versions match our lock files. Installing it would therefore give the user an environment corresponding to one that has been validated by CI.

The -deps-* packages were introduced to avoid duplicating the dependency specifications between firecrown and firecrown-validated. They are an software engineering choice to avoid having two places of maintenance for the same information, but are not strictly necessary.

marcpaterno:
@hmaarrfk I am not sure if a ping is necessary, or if you automatically get notified of my response (above); apologies if this is an unnecessary additional ping.

hmaarrfk:
I'm really not convinced that it is the right strategy. but i once said that if "globs" were not used, i wouldn't have "opinions" on the structure that developers choose. so i'll merge.

PR #2338, opened 2026-09-08
ADD: ocaml win-arm64

MementoRC:
## Checklist:

ping @conda-forge/ocaml
PR: TBD - experimental PR exists as proof of concept with zig, gcc/clang version in the work

### DIFF
--- a/requests/ocaml.yaml
+++ b/requests/ocaml.yaml
@@ -1,0 +1,4 @@
+action: add_feedstock_output
+feedstock_to_output_mapping:
+  ocaml:
+    - ocaml_win-arm64


carterbox:
This is a duplicate of #1864

MementoRC:
@carterbox It is not, one is to add a `-devel`, this one is to add the `win-arm64` output - Do you mean that they should be combined?
#1864 looked like it would be pushed-back strongly against, thus I did not bundle `win-arm64` into it

MementoRC:
@carterbox I think I understand, I linked the wrong PR in the `ocaml-devel` request because at the tie I had eventually bundled everything together - ocaml-devel request has no PR associated with it, the whole proposal is in the PR description, it would be just a matter of uncommenting, I can make a dedicated PR for it

This is the current PR for `win-arm64`: https://github.com/conda-forge/ocaml-feedstock/pull/103
It is also under re-write to make it more reviewable

carterbox:
It's a duplicate because they both request to add an output named: `ocaml_win-arm64`. I looked at the content of the PR not the title to decide that they were duplicates.

MementoRC:
My mistake.
I removed the duplication in `ocaml-devel`

MementoRC:
Thank you and apologies for the time spent from my sloppiness

PR #2341, opened 2026-09-11
Add the outputs of the smspp-project feedstock

dmeoli:
The smspp-project feedstock is being split into one package per module and per tool, in conda-forge/smspp-project-feedstock#26, so it needs the new output names registered: `libsmspp` and `libsmspp-<module>` for the libraries, `smspp-<module>` and `smspp-tools` for the command-line tools. `smspp-project` stays as the package installing everything, and none of the new names is taken on conda-forge.

cc @conda-forge/smspp-project

### DIFF
--- a/requests/smspp-project-outputs.yml
+++ b/requests/smspp-project-outputs.yml
@@ -1,0 +1,38 @@
+action: add_feedstock_output
+feedstock_to_output_mapping:
+  smspp-project:
+    - libsmspp
+    - libsmspp-bds
+    - libsmspp-bkb
+    - libsmspp-bnx
+    - libsmspp-bundle
+    - libsmspp-cflb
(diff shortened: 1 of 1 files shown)


carterbox:
We prefer if you mention the team @conda-forge/smspp-project instead of individuals, so that as reviewers, we don't have to check that the individual you have mentioned is actually a maintainer of the feedstock.

carterbox:
Please undraft when https://github.com/conda-forge/smspp-project-feedstock/pull/26 reaches a ready to merge state.

dmeoli:
Out of draft: conda-forge/smspp-project-feedstock#26 builds, tests and packages all 36 outputs on Linux, macOS and Windows, and the only failure left on every job is this validation.

PR #2357, opened 2026-09-21
chore: mark conda-smithy 20206.9.{21,22}

beckermr:
Native builds on riscv64 for v1 recipes do not work.

* You can use `pixi run find-filenames {matchspec}` to get a list of filenames matching given spec.

## Checklist:

cc @conda-forge/core


beckermr:
Native builds on riscv64 for linux are busted.

isuruf:
Let's not merge this until we know what's wrong and there's a conda-smithy issue.

h-vetinari:
> Let's not merge this until we know what's wrong and there's a conda-smithy issue.

https://github.com/conda-forge/conda-smithy/issues/2698

h-vetinari:
You can also pick some of your favourite riscv-enabled feedstocks and see for yourself if `provider: linux_riscv64: default` works. That's how I encountered all reported issues so far.

